PR #7, opened 2016-03-01
system-and-security-architecture.rst: remove TODO keyword

pohly:
Having "TODO" in the documentation was called out as a release
blocker. Let's use this PR to discuss how we can handle that properly.

[skip ci]

### DIFF
--- a/doc/architecture/system-and-security-architecture.rst
+++ b/doc/architecture/system-and-security-architecture.rst
@@ -235,17 +235,13 @@ users. It is not possible to set a root password.
 
 To become root in the core system:
 
-* After installation and before booting for the first time, add a
-  public key to the ``~root/.ssh/authorized_keys`` file (*TODO*:
-  create the directory and file with correct permissions, document
-  the exact location, which may vary between development and
-  production image)
+* Add a personal public key to the ``~root/.ssh/authorized_keys`` file,
+  then use ssh. There are different approaches for this (build-time
+  configuration option, modifying pre-built images, customizing a
+  running image) which are described in detail elsewhere.
 
-* \*Only in the development image\*: log in via a local console and or
-  serial port as root. A PAM module allows root to log in without
-  password. Because development and production image use different
-  signing keys (*TODO*), that module and its configuration cannot be
-  copied from a development image to a production image.
+* \*Only in the development image\*: root automatically gets logged in
+  on a local console or serial port.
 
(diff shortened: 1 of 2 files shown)


ipuustin:
I think this looks good. I would perhaps merge the two last commits. I also think that the TODOs make sense for the features that are not present yet in the distribution. Other than the TODO items, the documentation should reflect the current distribution state in the same repository.

kad:
I like usage of TODO as a keyword. Easy to search for things that shall be written.
It gives impression that "we care about this thing, but not yet done", rather than "nobody cares about this missing bit/aspect/corner case".

pohly:
On Tue, 2016-03-01 at 00:24 -0800, Ismo Puustinen wrote:
> I also think that the TODOs make sense for the features that are not
> present yet in the distribution.

So would you keep those TODO items in the documentation?

pohly:
On Tue, 2016-03-01 at 00:24 -0800, Ismo Puustinen wrote:
> I think this looks good. I would perhaps merge the two last commits.

I have intentionally not done that because the "remove resolved TODO
items" can be accepted permanently (items are resolved after all),
whereas "avoid using TODO keyword" is just (temporarily?) avoiding the
keyword for the release. We might accept that change and then revert it
again after the release.

kad:
Table of Content items, like below, I'd  keep to show potential readers that documentation is coming.
```
*TODO*: system provisioning and first boot security
```

For rest of TODO usages I have no any strong opinion.

dbkinder:
I helped Rajesh Sapra edit this document to remove TODOs for publishing to the website.

pohly:
Rebased onto current master.

I've reverted the changes introduced by PR #11 because they are partially incorrect. Some of my fixes were not included, so this PR still has them.

I removed the global TODO removal because these markers are useful. Let's merge this PR *after* the release and then rethink how we handle such markers going forward.

dbkinder:
We can merge this after the small tweaks we discussed on email are fixed.

kad:
@pohly @dbkinder so, what is the state of this PR ? if needed to be updated, let's update/rebase.

pohly:
Do not merge, this has to be rebased and needs further work - will do as soon as time permits.

pohly:
Revised commits pushed, @dbkinder please review.

mythi:
+1

dbkinder:
+1

PR #8, opened 2016-03-01
iptables: added a mechanism to override default firewall rules + update to the rule set

ipuustin:
To override the default iptables rules, you need to create your own custom firewall rule recipe that contains the firewall rule files. In this recipe you need to install your `iptables` and `ip6tables` rule files to the root filesystem and provide symbolic links to them from `/usr/share/iptables-settings/iptables.rules` and `/usr/share/iptables-settings/ip6tables.rules`. The default firewall rules are provided by `iptables-settings-default` package.

An example custom iptables rule recipe follows. The actual firewall rules are supposed to be in `files/custom-iptables.rules` and `files/custom-ip6tables.rules` files. The system integrator needs to make sure that the correct configuration package ends to the images, for example by setting `VIRTUAL-RUNTIME_iptables-settings` variable.

```
DESCRIPTION = "Custom iptables and ip6tables settings."
LICENSE = "MIT"
LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"

SRC_URI = " \
    file://custom-iptables.rules \
    file://custom-ip6tables.rules \
"

inherit update-alternatives

ALTERNATIVE_${PN} = "iptables.rules ip6tables.rules"
ALTERNATIVE_LINK_NAME[iptables.rules] = "/usr/share/iptables-settings/iptables.rules"
ALTERNATIVE_LINK_NAME[ip6tables.rules] = "/usr/share/iptables-settings/ip6tables.rules"

do_install() {
    install -d ${D}/usr/share/iptables-settings
    install -m 0644 ${WORKDIR}/custom-iptables.rules ${D}/usr/share/iptables-settings/iptables.rules

    if ${@bb.utils.contains('DISTRO_FEATURES', 'ipv6', 'true', 'false', d)}; then
        install -m 0644 ${WORKDIR}/custom-ip6tables.rules ${D}/usr/share/iptables-settings/ip6tables.rules
    fi
}
```


ipuustin:
Rebased to fix merge conflicts.

kad:
retest this please

ipuustin:
Force-pushed changes:
 * changed `/usr/share` to `${datadir}`
 * added `${datadir}/iptables-settings` to `FILES_${PN}` due to different semantics than `/etc/iptables-settings`
 * parameterized datadir in service files

ipuustin:
@pohly , @mythi : please review.

ipuustin:
@pohly: thanks for comments. I did the changes you requested. You were right -- the update-alternatives variables needed to be tweaked in order to allow the two packages to be installed at the same time. Force pushing now...

ipuustin:
@mythi : can you review and consider merging?

mythi:
@ipuustin I'm wondering the special code needed for IPv6. Would it be simpler to unconditionally add IPv6 too. IMO, it's almost de-facto for IoT devices. The sample code could add a note/comment saying ip6tables rules can be dropped in custom packaging if not needed.

kad:
:+1:

ipuustin:
@mythi : thanks for the comments! Implemented both of your suggestions.

klihub:
As a minimum requirement we have to allow DHCP and DNS request to come in from the container interfaces. Otherwise the container cannot even be bootstrapped to a working state. It could be done with service-specific ExecStartPre/ExecStopPost directives doing the necessary additions removals, but I think it would be just more painful without any actual additional security.
We also have to allow routing traffic from the container to the outside world. This might be something we don't want to unconditionally allow for all container with a single rule but rather make it a per container setting in the manifest. Needs more thinking to decide.

Anyway at least these should be added for now to the basic ruleset:

iptables -t filter -I INPUT -i ve-+ -p udp --dport 67 -j ACCEPT
iptables -t filter -I INPUT -i ve-+ -p udp --dport 53 -j ACCEPT

Additionally I would also add at least icmp (now unconditionally because the ICMP match module is not enabled in the kernel, if it was would just allow echo requests and replies) for the container interfaces:

iptables -t filter -I INPUT -i ve-+ -p icmp -j ACCEPT

Additionally I would seriously consider for now to unconditionally add:

iptables -t filter -I FORWARD -i ve-+ -j ACCEPT
iptables -t filter -I FORWARD -o ve-+ -j ACCEPT

to allow routing traffic from the containers forward and to allow transparent redirection of traffic to nspawned containers. We can turn these later into manifest-requested per-container rules, but I'd start with these common catch-all ones for now.

ipuustin:
@klihub: thanks for the comment! I'll add the virtual interface rule definitions to the default ruleset.

The ICMP ECHO support makes sense because IPv6 ping already works. There isn't any real security benefit in blocking  IPv4 ping but allowing IPv6 ping.

After merging this patch set we'll have the possibility to do different firewall settings packages. I think a good way forward would be to have a more permissive firewall in developer images, allowing basic network troubleshooting. In the production images I expect the ICMPv6 rules to be changed to a more granular direction and support for ping6 to disappear.

klihub:
Yes, that's probably a good idea. While talking about networking and development, I suggest we also add at least tcpdump to the development image variant. Can't really do any network debugging without useful tools...

pohly:
On Thu, 2016-03-10 at 06:57 -0800, Krisztian Litkey wrote:
> Yes, that's probably a good idea. While talking about networking and
> development, I suggest we also add at least tcpdump to the development
> image variant. Can't really do any network debugging without useful
> tools...

Agreed. This would fit nicely into the "interactive usage image feature"
that I added in my (to be merged) toybox branch. For example, iotop gets
added like this:

https://github.com/pohly/ostro-os/commit/0cb65dc33dff4cd718768f5693577a76cd455341

ipuustin:
@mythi : @dbkinder already went through the file and I did the changes he suggested. See https://github.com/ipuustin/meta-ostro/commit/69692b603b11223c2f627f9cd0222419f29f9a24 for the comments he made. The comment visibility was lost from the PR since the change was force pushed.

ipuustin:
@mythi: the build error appears not to be related to this PR.

mythi:
@ipuustin yes that is a know issue. it gets fixed with the new master.

mythi:
test this please

mythi:
Some unrelated build error. Restart one more time.

test this please

PR #10, opened 2016-03-01
swupd: Disable boot file heuristics

rojkov:
Ostro uses a mechanism different from what's used in ClearLinux
to detect changes in kernel, bootloader, initramfs.

This commit disables the CLR specific mechanism by commenting
out a call to boot file heuristics.


kad:
ok to test

rojkov:
I've added Upstream-Status tags to the patches lacking it and force pushed.

The reason why some patches were not submitted upstream is that we stick to swupd v5700 which was already old by the time the patches were developed. Updating swupd to the latest version might be quite a project.

rojkov:
It turned out some versions of `patch` may required leading whitespaces in context lines. I'll re-add the patches.

rojkov:
1. re-added the patches and rebased with --ignore-whitespace
2. changed Upstream-Status for "Fix build failure on Yocto" from Inappropriate to Submitted [https://github.com/clearlinux/swupd-client/pull/9]
3. force pushed

PR #19, opened 2016-03-02
Fix image creation for compressed dsk images, introduce xz compressed vdi images, introduce zip compressed images and update howto.

igor-stoppa:
@rojkov @okartau @Valtis @dbkinder  please help review

These patches do not alter what is currently produced by default, but if you are ok with them, I'll also submit a further one which leaves only compressed files.

### DIFF
--- a/meta-ostro/classes/image-dsk.bbclass
+++ b/meta-ostro/classes/image-dsk.bbclass
@@ -29,10 +29,19 @@
 IMAGE_TYPEDEP_dsk.xz = "dsk"
 IMAGE_TYPES_MASKED += "dsk_xz"
 
+IMAGE_TYPEDEP_dsk.zip = "dsk"
+IMAGE_TYPES_MASKED += "dsk.zip"
+
 IMAGE_TYPEDEP_dsk.vdi = "dsk"
 IMAGE_TYPES_MASKED += "dsk.vdi"
 
-IMAGE_DSK_ACTIVE = "${@bb.utils.contains_any('IMAGE_FSTYPES', 'dsk dsk_xz dsk.vdi', True, False, d)}"
+IMAGE_TYPEDEP_dsk.vdi_xz = "dsk"
+IMAGE_TYPES_MASKED += "dsk.vdi_xz"
+
+IMAGE_TYPEDEP_dsk.vdi.zip = "dsk"
+IMAGE_TYPES_MASKED += "dsk.vdi.zip"
+
+IMAGE_DSK_ACTIVE = "${@bb.utils.contains_any('IMAGE_FSTYPES', 'dsk dsk_xz dsk.zip dsk.vdi dsk.vdi_xz dsk.vdi.zip', True, False, d)}"
 
 do_uefiapp[depends] += " \
                          systemd:do_deploy \
@@ -50,6 +59,7 @@ IMAGE_DEPENDS_dsk += " \
                        dosfstools-native:do_populate_sysroot \
                        qemu-native:do_populate_sysroot \
                        xz-native:do_populate_sysroot \
+                       zip-native:do_populate_sysroot \
                      "
 INITRD_append = "${@ ('${DEPLOY_DIR_IMAGE}/' + d.getVar('INITRD_IMAGE', expand=True) + '-${MACHINE}.cpio.gz') if d.getVar('INITRD_IMAGE', True) and ${IMAGE_DSK_ACTIVE} else ''}"
 
--- a/doc/howtos/building-images.rst
+++ b/doc/howtos/building-images.rst
@@ -103,6 +103,39 @@ no IMA keys are needed::
     OSTRO_EXTRA_IMAGE_VARIANTS = "imagevariant:noima imagevariant:dev,noima"
 
 
+Image Formats
+-------------
+
+It is possible to produce different types of images:
+
+.dsk:
+    The basic format, which can be written to a block device with "dd".
+
+.dsk.vdi:
+    VirtualBox format, for running OSTRO inside a Virtual Machine.
+
+compressed formats:
+    Same as above, only compressed, to reduce (final) space occupation
+    and speed up the transfer between systems of the Ostro OS image.
+    Notice that the creation of compressed images will require additional
+    temporary space, because the creation of the compressed image depends
+    on the presence of the uncompressed one.
+
+To customize the image format, modify ``local.conf``, adding the variable
+``OSTRO_VM_IMAGE_TYPES``, set to any combination of the following::
+
+    dsk dsk_xz dsk.vdi dsk.vdi_xz
+
+It will also trigger the creation of corresponding symlinks.
+
+Example::
+
+    OSTRO_VM_IMAGE_TYPES = "dsk_xz dsk.vdi_xz"
+
+will create both the raw and the VirtualBox images, both compressed.
+
+
+
 Development Images
 ------------------
 
--- a/doc/howtos/building-images.rst
+++ b/doc/howtos/building-images.rst
@@ -122,11 +122,13 @@ compressed formats:
     Notice that the creation of compressed images will require additional
     temporary space, because the creation of the compressed image depends
     on the presence of the uncompressed one.
+    Compression is either based on xz (typical for Linux systems) or on zip
+    for operating systems that do not support xz natively.
 
 To customize the image format, modify ``local.conf``, adding the variable
 ``OSTRO_VM_IMAGE_TYPES``, set to any combination of the following::
 
-    dsk dsk_xz dsk.vdi dsk.vdi_xz
+    dsk dsk_xz dsk.zip dsk.vdi dsk.vdi_xz dsk.vdi.zip
 
 It will also trigger the creation of corresponding symlinks.
 
--- a/doc/howtos/building-images.rst
+++ b/doc/howtos/building-images.rst
@@ -103,8 +103,10 @@ no IMA keys are needed::
     OSTRO_EXTRA_IMAGE_VARIANTS = "imagevariant:noima imagevariant:dev,noima"
 
 
-Image Formats
--------------
+Image Formats for EFI platforms
+-------------------------------
+
+Note: The following chapter is applicable only to EFI platforms.
 
 It is possible to produce different types of images:
 
--- a/meta-ostro/lib/image-dsk.py
+++ b/meta-ostro/lib/image-dsk.py
@@ -177,13 +177,10 @@ def do_dsk_image():
 
     # If requested, create a xz compressed version of the dsk image.
     if "dsk_xz" in image_types:
-        xz_image_name = \
-            os.path.join(expand_vars("${DEPLOY_DIR_IMAGE}"),
-                         expand_vars('${IMAGE_NAME}.dsk.xz'))
+        check_call(['xz', '-3', '-vk', full_image_name])
         xz_image_name_link = \
             os.path.join(expand_vars("${DEPLOY_DIR_IMAGE}"),
                          expand_vars('${BPN}-${MACHINE}.dsk.xz'))
-        check_call(['xz', '-3', '-vk', full_image_name])
         symlink(expand_vars('${IMAGE_NAME}.dsk.xz'), xz_image_name_link)
     # If requested, create a zip compressed version of the dsk image.
     if "dsk.zip" in image_types:
--- a/meta-ostro/lib/image-dsk.py
+++ b/meta-ostro/lib/image-dsk.py
@@ -142,7 +142,8 @@ def do_dsk_image():
 
     image_types = lookup_var("IMAGE_FSTYPES").split()
     # Build .vdi images for use with VirtualBox, if so requested.
-    if "dsk.vdi" in image_types:
+    if set.intersection({"dsk.vdi", "dsk.vdi_xz", "dsk.vdi.zip"},
+                        set(image_types)) != set():
         vdi_image_name = \
             os.path.join(expand_vars("${DEPLOY_DIR_IMAGE}"),
                          expand_vars('${IMAGE_NAME}.dsk.vdi'))
@@ -151,7 +152,29 @@ def do_dsk_image():
                          expand_vars('${BPN}-${MACHINE}.dsk.vdi'))
         check_call(['qemu-img', 'convert', '-O', 'vdi',
                                 full_image_name, vdi_image_name])
-        symlink(expand_vars('${IMAGE_NAME}.dsk.vdi'), vdi_image_name_link)
+        # xz compressed vdi, if needed.
+        if "dsk.vdi_xz" in image_types:
+            check_call(['xz', '-3', '-vk', vdi_image_name])
+            vdi_xz_image_name_link = \
+                os.path.join(expand_vars("${DEPLOY_DIR_IMAGE}"),
+                             expand_vars('${BPN}-${MACHINE}.dsk.vdi.xz'))
+            symlink(expand_vars('${IMAGE_NAME}.dsk.vdi.xz'), vdi_xz_image_name_link)
+        # zip compressed vdi, if needed.
+        if "dsk.vdi.zip" in image_types:
+            check_call(['zip', '-9', vdi_image_name + ".zip", vdi_image_name])
+            vdi_zip_image_name_link = \
+                os.path.join(expand_vars("${DEPLOY_DIR_IMAGE}"),
+                             expand_vars('${BPN}-${MACHINE}.dsk.vdi.zip'))
+            symlink(expand_vars('${IMAGE_NAME}.dsk.vdi.zip'), vdi_zip_image_name_link)
+        # And / or uncompressed, if specified.
+        if "dsk.vdi" in image_types:
+            vdi_image_name_link = \
+                os.path.join(expand_vars("${DEPLOY_DIR_IMAGE}"),
+                             expand_vars('${BPN}-${MACHINE}.dsk.vdi'))
+            symlink(expand_vars('${IMAGE_NAME}.dsk.vdi'), vdi_image_name_link)
+        else:
+            os.remove(vdi_image_name)
+
     # If requested, create a xz compressed version of the dsk image.
     if "dsk_xz" in image_types:
         xz_image_name = \
@@ -162,6 +185,13 @@ def do_dsk_image():
                          expand_vars('${BPN}-${MACHINE}.dsk.xz'))
         check_call(['xz', '-3', '-vk', full_image_name])
         symlink(expand_vars('${IMAGE_NAME}.dsk.xz'), xz_image_name_link)
+    # If requested, create a zip compressed version of the dsk image.
+    if "dsk.zip" in image_types:
+        check_call(['zip', '-9', full_image_name + ".zip", full_image_name])
+        zip_image_name_link = \
+            os.path.join(expand_vars("${DEPLOY_DIR_IMAGE}"),
+                         expand_vars('${BPN}-${MACHINE}.dsk.zip'))
+        symlink(expand_vars('${IMAGE_NAME}.dsk.zip'), zip_image_name_link)
     # If the plain .dsk file was not requested, remove it and save space.
     if not "dsk" in image_types:
         os.remove(full_image_name)


kad:
ok to test

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -26,13 +26,16 @@
 
 # The mask used is dsk_xz instead of dsk.xz because of
 # https://bugzilla.yoctoproject.org/show_bug.cgi?id=9076
-IMAGE_TYPEDEP_dsk.xz = "dsk"
+IMAGE_TYPEDEP_dsk_xz = "dsk"
 IMAGE_TYPES_MASKED += "dsk_xz"
 
 IMAGE_TYPEDEP_dsk.vdi = "dsk"
```
kad: _dsk_vdi ?

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -26,13 +26,16 @@
...
 IMAGE_TYPES_MASKED += "dsk.vdi"
 
-IMAGE_DSK_ACTIVE = "${@bb.utils.contains_any('IMAGE_FSTYPES', 'dsk dsk_xz dsk.vdi', True, False, d)}"
+IMAGE_TYPEDEP_dsk.vdi_xz = "dsk"
```
kad: _dsk_vdi_xz ?

rojkov:
I'm ok with compressed only images: the current ci setup downloads *dsk.xz images for repo generation.

meta-ostro/classes/image-dsk.bbclass
igor-stoppa: No, the anomaly is the underscore, see the bug report I mentioned.
I'd rather not let it spread, and instead revert it back to a dot, once that bug is fixed.

pohly: Can you add a link to that bug to the commit? I haven't seen that bug mentioned.

pohly:
It's not clear to me from the commit how this will be used in practice. What are the IMAGE_FSTYPES values? Dot or underscore? What are the resulting file names?

Ideally, doc/howtos/building-images.rst should get updated to document this. Lacking that, provide enough information in the commit message such that @dbkinder can add such a section.

igor-stoppa:
@pohly 
Do you mean, apart from the line below?
+IMAGE_DSK_ACTIVE = "${@bb.utils.contains_any('IMAGE_FSTYPES', 'dsk dsk_xz dsk.vdi dsk.vdi_xz', True, False, d)}"

OK, I can add a comment to the ostro-image.bb file in a separate patch. But I'd prefer to create a separate PR, so that I can use the skip CI flag, since it is only documentation.

meta-ostro/classes/image-dsk.bbclass
igor-stoppa: to close this branch of the discussion, I pointed out on IRC that the link to the bug was already in the initial file, even if it didn't show up in the diff.

pohly:
Someone who knows where to look will of course be able to figure out what valid IMAGE_FSTYPES are, but not everyone will know that.

Instead of modifying the ostro-image.bb file, better directly update the howto. I'm not sure why a separate PR is better; IMHO including the documentation change here would help understand what the change is about and thus simplify the review.

But if that's what you prefer, then sure, do it like that.

igor-stoppa:
I was trying to avoid rebuilding only for a change in the docs.
OTOH I need to add support for .zip files, so I can put that patch here and add also the documentation change.

pohly:
Another, more fundamental question: can the existing support for compressing image files be used instead of working around it with dsk_xz instead of dsk.xz?

igor-stoppa:
IIRC it's not easily doable, because the image creation recipe makes assumptions about partition layout vs rootfs and the compression relies on those (the naming convention used for the image file).
The right way would be to fix the OE/YP recipe to be more generic, possibly using the dsk format, where the rootfs is just one of possibly many partitions (albeit special, since it's still the one mounted as root).
This, I suspect, won't be easy, because of the breakage it can and probably will create for existing users of that class/recipe.
It's also the reason why I intentionally went for a different extension, to make it obvious that it's something else and it's not guaranteed to be compatible with code that relies on .img files.

pohly:
I would have squashed some commits, but separate commits also works. I'm not fully convinced that we have to use a separate mechanism for adding compression, but don't have an alternative either, so overall, +1 from me.

pohly:
On Thu, 2016-03-03 at 02:43 -0800, Igor Stoppa wrote:
> IIRC it's not easily doable, because the image creation recipe makes
> assumptions about partition layout vs rootfs and the compression
> relies on those (the naming convention used for the image file).

I wasn't convinced that this is really hard to do, so I had a look
("seeing is believing"). The core problem is that the base classes add
".roofs" as suffix to the base image name, whereas in Ostro that suffix
was removed.

A patch making that suffix configurable in OE-core and thus allowing it
to be empty was easy. With that in place, quite a bit of code in Ostro
becomes redundant because we can use the compression and symlinking
support from OE-core.

I'll post that patch in
https://bugzilla.yoctoproject.org/show_bug.cgi?id=9076 to get feedback
from the OE-core devs. The corresponding change to Ostro is in
https://github.com/pohly/meta-ostro/tree/image-compression

We can merge the PR and remove the redundant code later (if desired), or
we can wait for feedback and then merge something else instead. I'm fine
with both approaches.

igor-stoppa:
On 4 March 2016 at 11:22, Patrick Ohly <notifications@github.com> wrote:
>
> I'll post that patch in
> https://bugzilla.yoctoproject.org/show_bug.cgi?id=9076 to get feedback
> from the OE-core devs. The corresponding change to Ostro is in
> https://github.com/pohly/meta-ostro/tree/image-compression
>
> We can merge the PR and remove the redundant code later (if desired), or
> we can wait for feedback and then merge something else instead. I'm fine
> with both approaches.


​The reason why I was thinking about it as hard task is that I was
considering to just drop the ".rootfs"​

​ part unconditionally and I was foreseeing problems with existing users.

So yes, if you make the dropping conditional to using a .dsk format, then
the problem disappears.

I still think that it's in incorrect to slap a ".rootfs" label to an image
that is not only a rootfs, but in the end it's a problem of those who stick
with that branch of the recipe and doesn't prevent us from using a proper
naming convention.


-- 
igor

pohly:
Is this PR here necessary to get builds working with current OE-core?

mythi:
@igor-stoppa are you ok to adapt image-dsk.* to the changes sent by @pohly to oe-core?

pohly:
On Mon, 2016-03-07 at 12:13 -0800, Mikko Ylinen wrote:
> @igor-stoppa are you ok to adapt image-dsk.* to the changes sent by
> @pohly to oe-core?

In case that you find it useful, the patch I was testing with is
currently here:
https://github.com/pohly/ostro-os/commit/79452288bb04b9f77a9402c4b0c36135bf1eaa0d

igor-stoppa:
@mythi As I wrote yesterday in the chat, there is also the appliance format coming, courtesy of @Valtis.
I would like to see first how that one shapes up, before moving to @pohly patches.
I also would like to see how they are received from oe-core.

Can't we get these merged and amend them afterwards?
I would like to transition CI to produce only compressed images ASAP.

mythi:
Coming back to this.

The status since Monday has changed. The proposed upstream changes are in master and master-next.

Let's move to use the upstream fixes and adapt our image creation to use those. Looks like @pohly has sample/test code ready so the work is expected to be smooth.

igor-stoppa:
The status has not changed.
To use the patches mentioned, I should introduce an incorrect - and custom to Ostro - handling of the vdi format. I still would prefer to do the transition only once, instead of having to create the hack that @pohly  has proposed.
The hack proposed by @pohly is that I should introduce a separate way of creating .vdi files, because what is provided by OE core works only for .hddimage files.
That will also introduce a misnomer, abusing the notion of compression with the .vdi format, which is not really a compression, it's a conversion.

kad:
Is this current content of PR conflicts with OE-core changes in pipeline ? if not, then let's get this one into our code, move forward and adopt to OE-core "better ways" via separate PR ?

igor-stoppa:
@kad No, there are no conflicts. The reason it is done as it is, is that it is intentionally written to avoid clashes with existing code.

kad:
then I'd suggest to move forward with improvements coming later as separate PR(s).

pohly:
I'd like to add that merging this PR and the proposed enhancements later implies changing user visible behavior, the name of the valid IMAGE_TYPES (current PR uses dsk_xz, future type will be dsk.xz).

@kad have you considered this impact on users (like getting error messages when they still use dsk_xz in their local.conf) when suggesting to merge this PR?

If you have and Mikko also agrees, then feel free to go ahead with merging. My position on this PR always has been that I wouldn't merge it myself, but will "disagree and commit" if others arrive at a different conclusion.

pohly:
Update on the situation in OE-core: the necessary patches are all in master now.

kad:
At the moment, there are no users for that feature. So, I expect if after one day or few days we would have better implementation with updates to documentation, it would be ok. Not a drastically big impact at the moment.

mythi:
On Wed, Mar 9, 2016 at 4:15 PM, Igor Stoppa wrote:

> The status has not changed.
> To use the patches mentioned, I should introduce an incorrect - and custom
> to Ostro - handling of the vdi format. I still would prefer to do the
> transition only once, instead of having to create the hack that @pohly
> <https://github.com/pohly> has proposed.
> The hack proposed by @pohly <https://github.com/pohly> is that I should
> introduce a separate way of creating .vdi files, because what is provided
> by OE core works only for .hddimage files.
> That will also introduce a misnomer, abusing the notion of compression
> with the .vdi format, which is not really a compression, it's a conversion.
>
> I'm not seeing a dependency to .hddimage. AFAICS the chained conversions
were applied on top of .dsk.

@igor-stoppa: are you saying even if this part we are now debating can be
cleaned up using the COMPRESSION* mechanism, you'd still have to implement
the appliance with the existing mechanism?

igor-stoppa:
@mythi you are not seeing the dependency because you are not looking in the right place.
What you should look at is the image-vm class, which implements *already* the support for vdi.
It only happens to do it for .hddimages, but it does it correctly.
What instead @pohly wants me to do is:
1) ignore that and implement in parallel another way to handle .vdi files
2) make the overloaded implementation inconsistent, because I should abuse the compression concept and sneak in through that the .vdi faking, which is a *conversion*, rather than a compression.

So, to summarize, this is the comment I'm getting:
"Hey, your code is not using the OE implementation and introduce an unclean underscore where a dot should be. Don't do THAT!!!11! Noooo! Instead use this other class which hides the problem and btw, while doing that, replicate other code and also introduce an inconsistency by troyan horsing a conversion through a compression. That looks sooooooo much better."

I beg to differ.

igor-stoppa:
@dbkinder No, the underscore vs dot is only inside the image selection string. File names and extensions have never been affected.

dbkinder:
Ah. Thanks for clarifying. I'm trying to keep an eye out for documentation changes...

mythi:
On Thu, Mar 10, 2016 at 12:56 AM, Igor Stoppa wrote:

> @mythi <https://github.com/mythi> you are not seeing the dependency
> because you are not looking in the right place.
> What you should look at is the image-vm class, which implements *already*
> the support for vdi.
> It only happens to do it for .hddimages, but it does it correctly.
> What instead @pohly <https://github.com/pohly> wants me to do is:
> 1) ignore that and implement in parallel another way to handle .vdi files
>
It's being ignored with current image-iot.bbclass too.

> 2) make the overloaded implementation inconsistent, because I should abuse
> the compression concept and sneak in through that the .vdi faking, which is
> a *conversion*, rather than a compression.
>
I understand this. With the latest patch the 'compression concept' becomes
a generic 'chained conversions' mechanism (it's just the variable
names remain COMPRESSION*).

> So, to summarize, this is the comment I'm getting:
> "Hey, your code is not using the OE implementation and introduce an
> unclean underscore where a dot should be. Don't do THAT!!!11! Noooo!
> Instead use this other class which hides the problem and btw, while doing
> that, replicate other code and also introduce an inconsistency by troyan
> horsing a conversion through a compression. That looks sooooooo much
> better."
>
> I beg to differ.
>

I think you are exaggerating. I'm seeing the chained conversion easier to
use/maintain from image-iot.bbclass perspective. However, I did not yet see
the answer to that appliance question. Is that in conflict with the
chaining stuff?

pohly:
On Wed, 2016-03-09 at 22:42 -0800, Mikko Ylinen wrote:
> I understand this. With the latest patch the 'compression concept' becomes
> a generic 'chained conversions' mechanism (it's just the variable
> names remain COMPRESSION*).

And just to be clear here, while it would be better to rename the
variables, doing so would be an API break that isn't possible right now
in OE-core. Even if it was, one would have to weigh the pain that this
will cause for external users against the benefits of better naming -
I'm not convinced that this is such a big problem.

It certainly doesn't cause any technical problems.

pohly:
On Wed, 2016-03-09 at 10:34 -0800, Alexander D. Kanevskiy wrote:
> At the moment, there are no users for that feature. So, I expect if
> after one day or few days we would have better implementation with
> updates to documentation, it would be ok. Not a drastically big impact
> at the moment.

Suppose that an implementation with IMAGE_FSTYPE = dsk.xz does not
appear within a few days. Do you still think we should merge the change
that introduces IMAGE_FSTYPES = dsk_xz as a temporary solution?

You had a nice quote the other day - "Nothing is more permanent than the
temporary" ;-}

igor-stoppa:
I prefer another QotD: Night brings Counsel

The problem here is that I disagree with what @mythi & @pohly want to do.
But it's neither technically difficult nor long to implement.
So I propose that they just put their name behind their words and submit the patches they would like to see. Then everyone is happy and we make actual progress.
I can change this PR to be only about docs.
Or whoever does the other implementation cna cherry pick the patches from here about docs.
As you prefer.

pohly:
On Thu, 2016-03-10 at 00:29 -0800, Igor Stoppa wrote:
> The problem here is that I disagree with what @mythi & @pohly want to
> do.
> But it's neither technically difficult nor long to implement.

Agreed. That's why I was suggesting it ;-}

> So I propose that they just put their name behind their words and
> submit the patches they would like to see. Then everyone is happy and
> we make actual progress.

Okay. I didn't want to do that initially because in the past you said
that you'd prefer to get enhancement suggestions during the code review
instead of having someone else fix your code after merging. But as you
explicitly gave your okay to have someone else propose an alternative
implementation and because both Mikko and Matti-Pekka also encouraged me
to do that, I'll prepare the necessary changes and submit a separate PR
with them.

Expect something today.

I'll also try to address some of your concerns about using code in
OE-core that still should be enhance further by including further
patches to OE-core that implement some of these enhancements. That'll
show that nothing in our code gets broken by these enhancements; I even
found a backward-compatible way to rename the
COMPRESSIONTYPES/COMPRESS_CMD/DEPENDS variables.

igor-stoppa:
On 11 March 2016 at 12:21, Patrick Ohly <notifications@github.com> wrote:

> On Thu, 2016-03-10 at 00:29 -0800, Igor Stoppa wrote:
> > The problem here is that I disagree with what @mythi & @pohly want to
> > do.
> > But it's neither technically difficult nor long to implement.
>
> Agreed. That's why I was suggesting it ;-}


​Indeed, it's just awful :-S​


-- 
igor

mythi:
On Fri, Mar 11, 2016 at 12:31 PM, Igor Stoppa <notifications@github.com>
wrote:

> On 11 March 2016 at 12:21, Patrick Ohly <notifications@github.com> wrote:
>
> > On Thu, 2016-03-10 at 00:29 -0800, Igor Stoppa wrote:
> > > The problem here is that I disagree with what @mythi & @pohly want to
> > > do.
> > > But it's neither technically difficult nor long to implement.
> >
> > Agreed. That's why I was suggesting it ;-}
>
>
> ​Indeed, it's just awful :-S​
>
>
Hey, don't judge too early :)

AFAIU, @pohly is looking to rename those
COMPRESSIONTYPES/COMPRESS_CMD/DEPENDS variables
in a backwards compatible way so the primary concern with .vdi coversion is
not compression should be gone.

igor-stoppa:
On 11 March 2016 at 13:21, Mikko Ylinen <notifications@github.com> wrote:

> On Fri, Mar 11, 2016 at 12:31 PM, Igor Stoppa <notifications@github.com>
> wrote:
>
> > On 11 March 2016 at 12:21, Patrick Ohly <notifications@github.com>
> wrote:
> >
> > > On Thu, 2016-03-10 at 00:29 -0800, Igor Stoppa wrote:
> > > > The problem here is that I disagree with what @mythi & @pohly want to
> > > > do.
> > > > But it's neither technically difficult nor long to implement.
> > >
> > > Agreed. That's why I was suggesting it ;-}
> >
> >
> > ​Indeed, it's just awful :-S​
> >
> >
> Hey, don't judge too early :)
>

​I judge based on the advice I was given and was also supported by you.
It was and still is awful.
​


> AFAIU, @pohly is looking to rename those
> COMPRESSIONTYPES/COMPRESS_CMD/DEPENDS variables
> in a backwards compatible way so the primary concern with .vdi coversion is
> not compression should be gone.


​If now there is some other way, fine, but it's not what I was asked to do.

Glad to see that also you guys are also not so eager to submit those hacks.​
Just wonder why you were advising me to do it ...


-- 
igor

pohly:
On Fri, 2016-03-11 at 03:27 -0800, Igor Stoppa wrote:
> Glad to see that also you guys are also not so eager to submit those
> hacks.​

I don't understand what you mean with "not so eager to submit" - I'm
doing some additional work to address your concerns (Yocto bug #9076),
and that of course takes time. The change in meta-ostro was trivial.

I also don't know what "those hacks" are, but I'm sure you will not
hesitate to call them out in the code.

Speaking of code, I've pushed all of my enhancements to:
https://github.com/pohly/ostro-os/tree/image-compression
and created a pull request:
https://github.com/ostroproject/ostro-os/pull/18

The additional commits for OE-core should address points 2 and 3 from
https://bugzilla.yoctoproject.org/show_bug.cgi?id=9076#c10

I have not worked on point 1 yet because a) we don't need it b) I'd like
to get some of the other changes reviewed first and c) I'm convinced
that it can be done fairly easily.

The actual change that we can merge into meta-ostro already now is here:
https://github.com/pohly/meta-ostro/tree/image-compression
https://github.com/ostroproject/meta-ostro/pull/42

PR #42 (lucky number!) replaces PR #19.

pohly:
> I have not worked on point 1 yet because a) we don't need it
> b) I'd like to get some of the other changes reviewed first and
> c) I'm convinced that it can be done fairly easily.

What the heck, after thinking about the problem I couldn't resist turning those thoughts into code - see "image.bbclass: prefer specialized image creation methods over chaining" in https://github.com/pohly/ostro-os/tree/image-compression.

I believe YOCTO #9076 can be considered resolved after merging all the commits prepared for it.

igor-stoppa:
I still think this is a hack.
Instead of introducing a sane and clean way to process modifiers progressively, it basically pulls the handbrake in case there is an alternative. A sort of poor man's method overloading.

What I would consider a clean implementation is:
* define one handler for each modifier
* apply the handlers sequentially, based on how modifiers are listed

mythi:
@igor-stoppa Hmm.. PR #42 implements what you think is a clean implementation (if the naming COMPRESSION vs. CONVERSION is ignored). How do you see it different?

igor-stoppa:
On 14 March 2016 at 13:27, Mikko Ylinen <notifications@github.com> wrote:

> @igor-stoppa <https://github.com/igor-stoppa> Hmm.. PR #42
> <https://github.com/ostroproject/meta-ostro/pull/42> implements what you
> think is a clean implementation (if the naming COMPRESSION vs. CONVERSION
> is ignored). How do you see it different?
>

​Yes, if we selectively ignore what is the core of my earlier comment,
plus we ignore that the workaround introduced makes it even more hackish,
it looks clean :-)​



-- 
igor

mythi:
Well the naming isn't relevant to the overall mechanism (which fulfills the criteria of 'clean implementation' above) . What is the workaround you're referring to?

igor-stoppa:
On 14 March 2016 at 14:12, Mikko Ylinen <notifications@github.com> wrote:

> Well the naming isn't relevant
>

​It is relevant in general​ - the misnomer is the first thing I pointed out.
You are entitled to have your own opinion, of course. Like I do.
Then you happen to be the one who merges. So be my guest, if you are ok
with it, but I still think it's wrong.

​We just traded one localized hack for a couple more distributed and spent
about 10 days in the process.

to the overall mechanism (which fulfills the criteria of 'clean
> implementation' above) . What is the workaround you're referring to?
>

​That it checks for the different implementation and then aborts the
default execution in favour of the exception., instead of having just one
implementation for each step.

Quoting:
----
For example, if there
 +    # was a fictional IMAGE_CMD_ext4.xz which creates a compressed
 +    # file directly, then we should use that instead of using
 +    # IMAGE_CMD_ext4 + CONVERSION_CMD_xz.
​----

Instead there simply should be one command for extension type(ext4, xz,
dsk, etc.)​ And they should be chained.

This implementation is trying to keep the foot in both camps.
I understand why, but it's not clean.
I had pointed it out initially, as explanation of why I didn't want to get
down this path. Now we finally have reached the point I had described.



-- 
igor

pohly:
On Mon, 2016-03-14 at 05:35 -0700, Igor Stoppa wrote:
> On 14 March 2016 at 14:12, Mikko Ylinen <notifications@github.com>
> wrote:
> 
> > Well the naming isn't relevant
> >
> 
> ​It is relevant in general​ - the misnomer is the first thing I
> pointed out.

And this will get address by future commits. We just can't do that now
because it would be too intrusive for OE-core, which is in preparation
of a release.

> > to the overall mechanism (which fulfills the criteria of 'clean
> > implementation' above) . What is the workaround you're referring to?
> >
> 
> ​That it checks for the different implementation and then aborts the
> default execution in favour of the exception., instead of having just
> one
> implementation for each step.

So you don't want "image.bbclass: prefer specialized image creation
methods over chaining" =
https://github.com/pohly/ostro-os/commit/be80d9330cccb496653d5c08b218b6e50e05b729 ?

I'm fine with that. As I said in
https://bugzilla.yoctoproject.org/show_bug.cgi?id=9076#c10 I consider
that nice to have at best, but not required.

But I had the impression that you did *not* want the core system to skip
over already defined image creation commands. In YOCTO #9076 you
reported as problem:

> when I tried to support also its xz-compressed version, represented
> with the type dsk.xz, this triggers a bogus match and the build fails
> with trying to compress a file that doesn't exist:
> IMAGE_NAME}.rootfs.${type}

The commit above fixes exactly that. I tried it by using the
image-dsk.bbclass with IMAGE_TYPE_MASKED = "dsk.xz" and the modified
image.bbclass worked as you asked it to behave:

> Ideally the image_types class would be more flexible, but, lacking
> that, it should not try to gobble file types that are not its
> business.

The class now is more flexible (we can use the builtin xz and vdi
conversion mechanism) and, if explicitly asked to do so, will not inject
the unwanted and non-functional xz compression code anymore.

> This implementation is trying to keep the foot in both camps.

Only because I thought you wanted that. I'm fine with dropping the
commit above. I can imagine some valid use cases (as described in the
commit message), but it's not a very strong argument for including the
change.

igor-stoppa:
On 14 March 2016 at 16:11, Patrick Ohly <notifications@github.com> wrote:

> On Mon, 2016-03-14 at 05:35 -0700, Igor Stoppa wrote:
> > On 14 March 2016 at 14:12, Mikko Ylinen <notifications@github.com>
> > wrote:
> >
> > > Well the naming isn't relevant
> > >
> >
> > ​It is relevant in general​ - the misnomer is the first thing I
> > pointed out.
>
> And this will get address by future commits. We just can't do that now
> because it would be too intrusive for OE-core, which is in preparation
> of a release.
>

​The same could have been said of the initial submission.​
But I see no reason why a patch should be timed against OE-Core.
It can be kept in Ostro and merged into OE-Core later on, when the merge
window is opened once more.


> > > to the overall mechanism (which fulfills the criteria of 'clean
> > > implementation' above) . What is the workaround you're referring to?
> > >
> >
> > ​That it checks for the different implementation and then aborts the
> > default execution in favour of the exception., instead of having just
> > one
> > implementation for each step.
>
> So you don't want "image.bbclass: prefer specialized image creation
> methods over chaining" =
>
> https://github.com/pohly/ostro-os/commit/be80d9330cccb496653d5c08b218b6e50e05b729
> ?
>
> I'm fine with that. As I said in
> https://bugzilla.yoctoproject.org/show_bug.cgi?id=9076#c10 I consider
> that nice to have at best, but not required.
>

​It's a hack because it leaves both ways enabled and supported.
​


> But I had the impression that you did *not* want the core system to skip
> over already defined image creation commands. In YOCTO #9076 you
> reported as problem:
>
> > when I tried to support also its xz-compressed version, represented
> > with the type dsk.xz, this triggers a bogus match and the build fails
> > with trying to compress a file that doesn't exist:
> > IMAGE_NAME}.rootfs.${type}
>

I was simply complaining about the match, which I consider incorrect.​


> The commit above fixes exactly that. I tried it by using the
> image-dsk.bbclass with IMAGE_TYPE_MASKED = "dsk.xz" and the modified
> image.bbclass worked as you asked it to behave:
>
> > Ideally the image_types class would be more flexible, but, lacking
> > that, it should not try to gobble file types that are not its
> > business.
>

​Exactly what I said. It's not its business.​


> The class now is more flexible (we can use the builtin xz and vdi
> conversion mechanism) and, if explicitly asked to do so, will not inject
> the unwanted and non-functional xz compression code anymore.
>
> > This implementation is trying to keep the foot in both camps.
>
> Only because I thought you wanted that. I'm fine with dropping the
> commit above. I can imagine some valid use cases (as described in the
> commit message), but it's not a very strong argument for including the
> change.


​The problem is that the foot you want to preserve is in the wrong camp.

​What I already said would be the clean solution is:

* one handler for each extension​

​(dsk, ext4, xz, zip, etc)​
​* chain handlers in the order the extensions are used​


-- 
igor

pohly:
On Mon, 2016-03-14 at 07:35 -0700, Igor Stoppa wrote:
> ​The same could have been said of the initial submission.​
> But I see no reason why a patch should be timed against OE-Core.
> It can be kept in Ostro and merged into OE-Core later on, when the
> merge window is opened once more.

Except that we cannot keep patches against image.bbclass in Ostro. We
would have fork the entire class. Living with the COMPRESSION_CMD names
for now seems like the lesser evil.

> > So you don't want "image.bbclass: prefer specialized image creation
> > methods over chaining" =
> >
> >
> https://github.com/pohly/ostro-os/commit/be80d9330cccb496653d5c08b218b6e50e05b729
> > ?
> >
> > I'm fine with that.

Just to be sure that there is no misunderstanding: what I meant is, "I'm
fine with not wanting it", and had already said that it is not needed.

> ​It's a hack because it leaves both ways enabled and supported.

I take that as a "no, I don't want it". If we drop that, is anything
left in
https://github.com/pohly/ostro-os/tree/ffb0591b489f01d0b1eef2d0329142799625ec7e that you consider a hack?

> > The commit above fixes exactly that. I tried it by using the
> > image-dsk.bbclass with IMAGE_TYPE_MASKED = "dsk.xz" and the modified
> > image.bbclass worked as you asked it to behave:
> >
> > > Ideally the image_types class would be more flexible, but, lacking
> > > that, it should not try to gobble file types that are not its
> > > business.
> >
> 
> ​Exactly what I said. It's not its business.​

Except that there is a COMPRESS_CMD_xz, so image file types ending
in .xz are, by definition, the business of the image.bbclass. So I'm
still unsure how you wanted YOCTO #9076 to be fixed. At which point can
it be closed?

> ​What I already said would be the clean solution is:
> 
> * one handler for each extension​
> 
> ​(dsk, ext4, xz, zip, etc)​
> ​* chain handlers in the order the extensions are used​

Okay, I think we have that when dropping that last commit. Agreed?

mythi:
Superseded by #42 so closing.

We can still finish the conversation on which commits in https://github.com/pohly/ostro-os/commits/image-compression are needed and which are not.

PR #22, opened 2016-03-04
systemd: port gummiboot cmdline patch

pohly:
Systemd 2.29 added some more code to the if() body that the patch
removes.

Beware that merging this makes meta-ostro unusable unless we also merge OE-core master!
Do not merge yet unless all pieces are in place.

[skip ci]


kad:
@pohly can I suggest to make test PR to ostro-os with only selected fixes that blocks OE-core ?
this way we can check that it works without skipping CI test

pohly:
On Fri, 2016-03-04 at 23:51 -0800, Alexander D. Kanevskiy wrote:
> @pohly can I suggest to make test PR to ostro-os with only selected
> fixes that blocks OE-core ?
> this way we can check that it works without skipping CI test

I can do that once we have the last missing piece, the ported systemd
patches in meta-appfw.

mythi:
@kad FYI The test build is finally ready and shows blue: https://github.com/ostroproject/ostro-os/pull/7

mythi:
pushed in master.

PR #25, opened 2016-03-07
packagegroup-devkit: add soletta-dev-app to be shipped to dev images

brunobottazzini:
Adds Soletta Dev-App to be installed in images that contains packagroup-devkit.
This package group contains IoT development kit related components.

If the image is for production it should not contain the Dev-App, since it is
an IDE for development purposes.


mythi:
-1

devkit packagegroup is only for iot-devkit middleware packages and should be kept independent. The best match would be to add soletta-dev-app in ostro-imag-dev image directly (since it already contains the necessary build tools and utils like git). Another option would be to bake 'soletta-dev' imagevariant.

brunobottazzini:
Thanks i will make a V2

PR #26, opened 2016-03-07
DOC: clarify dd command in booting-and-installation doc

dbkinder:
Clarified you should write image to the whole SD card device (of=/dev/mmcblk0) and not
to just a partition (of=/dev/mmcblk0p1) - user feedback

[skip ci]

### DIFF
--- a/doc/howtos/booting-and-installation.rst
+++ b/doc/howtos/booting-and-installation.rst
@@ -32,7 +32,7 @@ Using dd to Create Bootable Media
 
 Once you have the :file:`.dsk.xz` Ostro OS image you need to get it
 onto your hardware platform, typically by using removable media such as a 
-USB thumb drive or SD card.  The typical way to do this is with the :command:`dd` command.
+USB thumb drive or SD card.  The usual way to do this is with the :command:`dd` command.
 
 #. Connect your USB thumb drive or SD card to your Linux-based development system
    (minimum 8 GB card required).
@@ -41,13 +41,18 @@ USB thumb drive or SD card.  The typical way to do this is with the :command:`dd
 
      $ dmesg 
 
-   or you can use the :command:`lsblk` command to show the block-level devices; a USB drive usually shows up as ``/sdb`` or ``/sdc``
-   (almost never as ``/sda``)
+   or you can use the :command:`lsblk` command to show the block-level devices; a USB drive usually 
+   shows up as ``/sdb`` or ``/sdc``
+   (almost never as ``/sda``) and an SD card would usually show up as :file:`/dev/mmcblk0`.  
+   
+   Note: You should specify the whole device you're writing to with 
+   :command:`dd`:  (e.g., :file:`/dev/sdb` or
+   :file:`/dev/mmcblk0`) and **not** just a partition on that device (e.g., :file:`/dev/sdb1` or
+   :file:`/dev/mmcblk0p1`) on that device. 
 
 #. The :command:`dd` command will overwrite all content on the device so be careful specifying 
    the correct media device. In the example below, :file:`/dev/sdb` is the 
-   destination USB device on our development machine (an SD card device would typically show up 
-   as :file:`/dev/mmcblkX` where `X` is a number)::
+   destination USB device on our development machine::
 
       $ sudo umount /dev/sdb*
       $ xzcat <ostro-os-image.dsk.xz> | sudo dd of=/dev/sdb bs=512k


doc/howtos/booting-and-installation.rst
```diff
@@ -41,13 +41,18 @@ USB thumb drive or SD card.  The typical way to do this is with the :command:`dd
...
 
-   or you can use the :command:`lsblk` command to show the block-level devices; a USB drive usually shows up as ``/sdb`` or ``/sdc``
-   (almost never as ``/sda``)
+   or you can use the :command:`lsblk` command to show the block-level devices; a USB drive usually 
+   shows up as ``/sdb`` or ``/sdc``
+   (almost never as ``/sda``) and an SD card would usually show up as :file:`/dev/mmcblk0`.  
+   
+   Note: You should specify the whole device you're writing to with 
+   :command:`dd`:  (e.g., :file:`/dev/sdb` or
+   :file:`/dev/mmcblk0`) and **not** just a partition on that device (e.g., :file:`/dev/sdb0` or
```
kad: sdb0 impossible combination. it might be sdb1, sdb2,...

dbkinder:
fixed.

kad:
:+1:

mythi:
+1

doc/howtos/booting-and-installation.rst
```diff
@@ -41,13 +41,18 @@ USB thumb drive or SD card.  The typical way to do this is with the :command:`dd
...
-   or you can use the :command:`lsblk` command to show the block-level devices; a USB drive usually shows up as ``/sdb`` or ``/sdc``
-   (almost never as ``/sda``)
+   or you can use the :command:`lsblk` command to show the block-level devices; a USB drive usually 
+   shows up as ``/sdb`` or ``/sdc``
+   (almost never as ``/sda``) and an SD card would usually show up as :file:`/dev/mmcblk0`.  
+   
+   Note: You should specify the whole device you're writing to with 
+   :command:`dd`:  (e.g., :file:`/dev/sdb` or
+   :file:`/dev/mmcblk0`) and **not** just a partition on that device (e.g., :file:`/dev/sdb1` or
+   :file:`/dev/mmcblk0p1`) on that device.
```
barbieri: I know this was merged, so my recommendation goes to another commit to improve this one. Please consider using full examples or even better let's provide a script. This looks reasonable:

Human readable instructions:
```sh
# check the required uncompressed space
user@host# xz -l ostro.dsk.xz 
Strms  Blocks   Compressed Uncompressed  Ratio  Check   Filename
    1       1     86.8 MiB     87.3 MiB  0.994  CRC64   ostro.dsk.xz

# list available block devices, pick the best that works for you. We need XXGb of space
user@host$ lsblk --list --paths --nodeps --output NAME,HOTPLUG,SIZE
NAME     HOTPLUG        SIZE
/dev/sda       0 32G
/dev/mmcblk0 1 8G

# say you want /dev/mmcblk0

# confirm if it's not mounted (in this case shows no output):
user@host$ findmnt --list --noheading --output SOURCE,TARGET | grep '^/dev/mmcblk0'

# if command above shows no matches, call dd to write image:
user@host$ xzcat ostro.dsk.xz | sudo dd of=/dev/mmcblk0 bs=512K
```

Automated script (host at ostro-os/scripts?): **UNTESTED**
```sh
#!/bin/bash

IMG=${1:?missing image.dsk.xz See https://download.ostroproject.org/releases/ostro-os/}
DEV=${2}

die() {
    echo "ERROR: $*" >&2
    exit 1
}

for c in lsblk findmnt xz; do
    if ! which $c; then
        die "missing '$c' command line utility."
    fi
done

IMG_SIZE=`xz --robot -l "$IMG" | grep -e '^totals'  | cut -f5`

if [ -z "$DEV" ]; then
    echo "NOTE: no device given, available devices are (only hot-pluggable devices displayed):"
    lsblk --paths --list --nodeps --noheadings  --output HOTPLUG,NAME,SIZE
     #| grep -e '^ *1 '
    echo -n "Your choice: "
    read DEV
fi

DEV_SIZE=`lsblk ${DEV} --bytes --list --nodeps --output SIZE | head -n1`

if [ ${DEV_SIZE} -lt ${IMG_SIZE} ]; then
    die "disk ${DEV} size ${DEV_SIZE} bytes is not large enough to fit uncompressed ${IMG} of ${IMG_SIZE} bytes"
fi

MNTS=`findmnt --list --noheading --output SOURCE,TARGET | sed -e 's:^\([^ ]\+\) \+\([^ ]\+\):\1=\2:g' | grep -e "^${DEV}"`
if [ -n "$MNTS" ]; then
    die "disk ${DEV} contains mountpoints: ${MNTS}"
fi

echo "writing ${IMG} to ${DEV}: ${IMG_SIZE} bytes"

xz --verbose --decompress --stdout "${IMG}" | sudo dd of="${DEV}" bs=4M status=progress conv=nocreat,notrunc,fdatasync

if which eject >/dev/null 2>/dev/null; then
    sudo eject ${DEV}
fi
```

gvancuts:
I like the direction this is going although I would go one step further and try to provide a GUI. To me, this nearly sounds like an engineering deliverable that we would eventually document. Or are you expecting the TME team to own and maintain such script?

barbieri:
I think engineering should write and maintain the script.

I would love to see a GUI as well, recommendations?

One remark that maybe lead us to keep both is that this script would work on all Linux system due the tools used, while GUI would need things like pygtk or pyqt (or be natively compiled) and likely a package for each distro :-/

However, if we do have a nice GUI we can document and maintain both it and the script.

By the way, it would be even better to offer a way to automatically download images :-)

gvancuts:
I don't really have any recommendation for a GUI, being able to flash from your dev environment would be nice (e.g. Eclipse) and I also very much like the idea of being able to automatically download the image from the download.ostroproject.org website. We may need to do this in several iterations though.

I was thinking a UNetBootin for a short while because it runs a several OS and can download images but I'm not sure you can do a raw copy of an image with it, I think it always extracts the image and installs it which would not work for us.

barbieri:
btw, @gvancuts do we have instructions on how to do it on windows? We just had someone trying to use it for Soletta and was on windows... major nightmare for us :-)

dbkinder:
All our M3 documentation assumes a Linux development environment (not Mac or Windows), primarily because Yocto Project tools only run on Linux.  That said, I could add Windows instructions for using, for example https://sourceforge.net/projects/win32diskimager/ to get the image on bootable media.

gvancuts:
+1 on adding instructions for flashing from Mac and Windows. I've seen that request too already (and I do believe that using Win32 Disk Imager works fine)

PR #27, opened 2016-03-08
DOC: add use of SSTATE_MIRRORS to building-images doc

dbkinder:
- Added information about enabling SSTATE_MIRRORS (and references to Yocto Project
  documentation for details)
- Updated platform/build target mapping to include specific platforms

[skip ci]

### DIFF
--- a/doc/howtos/building-images.rst
+++ b/doc/howtos/building-images.rst
@@ -47,9 +47,9 @@ these keys or disable the features which depend on them.
 
 In addition, images are locked down by default: for example, none of
 the existing user accounts (including root) has a password set, so
-logging into the running system is impossible. Some way of interacting
-with the system after booting it has to be chosen before building
-images.
+logging into the running system is impossible. Before building an image,
+you must choose a way of interacting with the system after it has booted.
+
 
 Target MACHINE Architecture
 ----------------------------
@@ -62,14 +62,14 @@ For currently :ref:`platforms`, the appropriate ``MACHINE`` selections are:
 
 .. table:: Yocto MACHINE selection for Supported Hardware platforms
 
-    ============  ====================================
-    Platform      Yocto Project MACHINE selection
-    ============  ====================================
-    GigaByte      intel-corei7-64
-    Galileo       intel-quark
-    MinnowBoard   intel-corei7-64
-    Edison        edison
-    ============  ====================================
+    ==========================  ====================================
+    Platform                    Yocto Project MACHINE selection
+    ==========================  ====================================
+    GigaByte GB-BXBT-3825       intel-corei7-64
+    Intel Galileo Gen2          intel-quark
+    MinnowBoard MAX compatible  intel-corei7-64
+    Intel Edison                edison
+    ==========================  ====================================
 
 Virtual machine images (a :file:`.vdi` file) are created for each of these builds hardware platforms as part 
 of the build process (and included in the prebuilt image folders too).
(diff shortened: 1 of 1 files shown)


mythi:
+1

mythi:
@dbkinder for the future PRs, it'd good to have the independent edits separated in their own commits. (My preference.) See, e.g. #7

doc/howtos/building-images.rst
```diff
@@ -122,6 +122,26 @@ with ``# require conf/distro/include/ostro-os-development.inc`` and
...
+parts do not need to be rebuilt. The Yocto Project shared state code supports
+incremental builds and attempts to accelerate build time through the use
+of prebuilt data cache objects configured with the ``SSTATE_MIRRORS`` setting.
+
+By default, this ``SSTATE_MIRRORS`` configuration is disabled in :file:`conf/local.conf`
+but can be easily enabled by uncommenting the ``SSTATE_MIRRORS`` line
+in your :file:`conf/local.conf` file, as shown here:::
+
+   # Example for Ostro OS setup, recommended to use it:
+   #SSTATE_MIRRORS ?= "file://.* http://download.ostroproject.org/sstate/ostro-os/PATH"
```
barbieri: @mythi @kad wasn't it the goal to enable this by default?

kad: it was, but for some reason become forgotten :(

mythi: I'll fix it once I get the big queue empty.

dbkinder: If/when you do change the default setting for this, please update the documentation as well :)   Do you think you'd want to let developers know how to disable use of SSTATE_MIRRORS if we change the default to have it enabled?

barbieri: Likely not, but it is worth to keep a mention what is SSTATE, it's the core to deliver the precompiled binaries.

pohly: On Thu, 2016-03-10 at 04:50 -0800, Gustavo Sverzut Barbieri wrote:
> Likely not, but it is worth to keep a mention what is SSTATE, it's the
> core to deliver the precompiled binaries. 

But isn't SSTATE mirror as described here just a stop-gap measure? Just
configuring the mirrors suffers from poorer utilization of the SSTATE
because native tools still depend a lot on the host distribution.

eSDK will elevate that to some extend by running pre-compiled native
tools with a custom libc and thus avoid some of the host dependencies.

barbieri: eSDK AFAIU also uses SSTATE for othe bite, just the native tools comes from different path, coming in the SDK itself and is modified to be relocatable and host-agnostic 

How to solve the host-agnostic coming from SSTATE is still to be solved, @kad is talking to Yocto folks.

pohly: On Thu, 2016-03-10 at 05:30 -0800, Gustavo Sverzut Barbieri wrote:
> eSDK AFAIU also uses SSTATE for othe bite, just the native tools comes
> from different path, coming in the SDK itself and is modified to be
> relocatable and host-agnostic

True. But my point was more about our documentation: do we still need to
document SSTATE_MIRRORS and how to use them directly? Is that still
relevant once we start using the eSDK?

> How to solve the host-agnostic coming from SSTATE is still to be
> solved, @kad is talking to Yocto folks. 

eSDK does something regarding that with uninative.bbclass, but
apparently there are still issues left.

PR #31, opened 2016-03-08
ostro-image.bb: creates soletta-dev packagegroup - v2

brunobottazzini:
Cretes a packagegroup that contains the development packages for Soletta.

This packages is only needed for dev images and not production images, since
the packages should be added is only for development purposes.

Differences from V1:
 - As @mythi suggested, it shouldn't be added in devkit and we should create a soletta-dev feature. 
   
Since we would add more feature in the future I created a packagegroup that we can add the dev   packages for Soletta

Previous patch:
https://github.com/ostroproject/meta-ostro/pull/25

### DIFF
--- a/meta-ostro/recipes-core/packagegroups/packagegroup-soletta-dev.bb
+++ b/meta-ostro/recipes-core/packagegroups/packagegroup-soletta-dev.bb
@@ -1,0 +1,8 @@
+SUMMARY = "Development Packages for Soletta IoT Framework"
+LICENSE = "MIT"
+
+inherit packagegroup
+
+RDEPENDS_${PN} = " \
+    soletta-dev-app \
+"
--- a/meta-ostro/recipes-image/images/ostro-image.bb
+++ b/meta-ostro/recipes-image/images/ostro-image.bb
@@ -29,6 +29,7 @@ IMAGE_FEATURES[validitems] += " \
     qatests \
     smack \
     soletta \
+    soletta-dev \
     swupd \
     tools-develop \
 "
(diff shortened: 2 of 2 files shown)


mythi:
@brunobottazzini the change itself looks OK but please check the CI error this PR gives.

I'm not sure what the problem is...

brunobottazzini:
@mythi, I looked at the CI to investigate the problem and found this:

http://pastebin.intel.com/view/b396426d

I had this problem once with yocto and I had to clean my tmp.

I tested here building manually and it worked normally.

Do you think we can clean this tmp to see if this will work ? Or do you think this might be another problem ?

kad:
@brunobottazzini all builds in OstroProjectCI are done in clean directory, so there is no tmp present when build starts.

barbieri:
please use public pastebin, this is a publi project now :-)

kad:
@brunobottazzini @mythi ycoto tools automatically for each group creating names package-group-name-{dev,debug,ptest}. thus new package name packagegroup-soletta-dev from this PR would conflict with automatically created packagegroup-soletta-dev from packagegroup-soletta. This is the reason of build failure.

brunobottazzini:
@kad, Interesting, it makes sense somehow I didn't get this error here. Maybe if I start a clean build the error happens.

I will make a V3 with a different name for packagegroup-soletta-dev then

mythi:
@kad Thanks, makes sense (btw, perhaps the package-group bbclass should disable that) I was suspecting a naming conflict but could not come up with the evidence. :)

PR #32, opened 2016-03-08
ostro.conf: set linux-yocto preferred version to 4.1

mythi:
oe-core/meta-intel recently moved to 4.4 on all MACHINEs we are using.

However, some of the Ostro patches that are applied via linux-yocto_%.bbappend
in meta-ostro-bsp are not (yet) compatible with 4.4. The build fails with
MACHINEs we have not properly set PREFERRED_VERSION to 4.1 for linux-yocto.

Make sure all MACHINEs set PREFERRED_VERSION to 4.1 until we are
ready to move to 4.4.

### DIFF
--- a/meta-ostro/conf/distro/ostro.conf
+++ b/meta-ostro/conf/distro/ostro.conf
@@ -58,10 +58,10 @@ DISTRO_FEATURES_DEFAULT_remove = "x11 3g"
 
 DISTRO_FEATURES ?= "${DISTRO_FEATURES_DEFAULT} ${DISTRO_FEATURES_LIBC} ${OSTRO_DEFAULT_DISTRO_FEATURES}"
 
-# Use 4.1 kernel for core2 and corei7.
-# BeagleBone (and others) remain unchanged on default kernel version.
-PREFERRED_VERSION_linux-yocto_intel-core2-32 ?= "4.1%"
+# Use 4.1 kernel for all MACHINEs
 PREFERRED_VERSION_linux-yocto_intel-corei7-64 ?= "4.1%"
+PREFERRED_VERSION_linux-yocto_intel-quark ?= "4.1%"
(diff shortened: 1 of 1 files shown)


mythi:
@jlaako Please check.

jlaako:
Looks good, should work fine with the upstream meta-intel, since it has both 4.1 and 4.4 kernels available.

mythi:
rebased + force pushed

pohly:
test this please

The previous test failed on Gigabyte: https://ostroproject.org/jenkins/job/test_gigabyte/102/
Let's retest.

mythi:
@pohly thanks

pohly:
Now it failed on Galileo instead: https://ostroproject.org/jenkins/job/test_galileov2/106/console

The patch itself  looks fine. If these are spurious test failures that also occur elsewhere, then we should merge it.

kad:
failed reboot test might be indicator of problem within kernel. better to investigate. @Valtis

pohly:
On Thu, 2016-03-10 at 01:23 -0800, Alexander D. Kanevskiy wrote:
> failed reboot test might be indicator of problem within kernel. better
> to investigate. @Valtis

Agreed. However, if it also occurs in other PRs, then it is not caused
by this one here and thus this change then should be okay to merge.
That's why I asked whether it is a know problem.

mythi:
FWIW, galileo2 test also fails with an 'edison-only' change in meta-ostro-bsp#7

mythi:
@pohly @kad OK to merge?

Valtis:
As far as I know, Galileo has either firmware or kernel bug where it randomly fails to reboot, so random reboot failures are to be expected.

pohly:
@mythi yes, I'm fine with merging.

PR #34, opened 2016-03-09
README: "iot layer" -> "meta-ostro layer".

ipuustin:
[skip ci]

The README should be about "meta-ostro" layer.


mythi:
nice catch!

PR #35, opened 2016-03-09
ostro-image.bb: creates soletta-dev-tools packagegroup V3

brunobottazzini:
ostro-image.bb: creates soletta-dev-tools packagegroup

Cretes a packagegroup that contains development tools for Soletta.

Since these tools are only for development purposes it needs to be included
only in dev images.

Differences from V2:
As pointed from @kad and @mythi, we had a name conflict and the build was failing. 
Changed the name to be soletta-dev-tools instead of soletta-dev.

Differences from V1:

As @mythi suggested, it shouldn't be added in devkit and we should create a soletta-dev feature.
Since we would add more feature in the future I created a packagegroup that we can add the dev packages for Soletta

Previous patch:
https://github.com/ostroproject/meta-ostro/pull/31


barbieri:
You can call it only soletta-tools, since sockets is a development platform there is no need to call it -dev-tools

barbieri:
btw, do we really need a package group? IMHO the devapp package will depend on everything, so there is no need to have s group for it.

brunobottazzini:
@barbieri, soletta-tools is a good name. I'll change it. 

You're right we do not need the packagegroup because there is only soletta-dev-app. 

However, @bdilly said to me he'd like to include more things in the future. So I created the packagegroup to include these set of tools there.

kad:
@barbieri package group is better because if we want to have same set of features included in multiple images, it's easier to copy only one line, than maintaining multiple copies of linger lists.

barbieri:
@kad my point is that devapp (webui node.js) itself will do that work by depending on all it needs, then there is no need for packagegroup.

@bdilly could you clarify what would not?

mythi:
@barbieri IIRC soletta-dev-app dependencies are set up like that (e.g., it RDEPENDS on git). 

FWIW, it is also possible to set the IMAGE_FEATURES by simply
```
FEATURE_PACKAGES_soletta-dev-tools = "soletta-dev-app"
```

If we want to make a compromise, we can start with that and add the packagegroup later if there's a need.

brunobottazzini:
@mythi, indeed, we can postpone the packagegroup when there will be the need. :+1:

PR #36, opened 2016-03-09
Switch swupd to use bsdtar instead of GNU tar

rojkov:
Currently GNU tar can't be used when IMA is enabled because
upon extracting a file it first creates it with `mknodat()`,
sets its xattrs and only then tries to open for writing and
fails since IMA prevents it.

This patch adds patches to swupd making it compatible with
bsdtar and switches it to actually use bsdtar.

### DIFF
--- a/meta-ostro/recipes-swupd/packagegroups/packagegroup-swupd.bb
+++ b/meta-ostro/recipes-swupd/packagegroups/packagegroup-swupd.bb
@@ -7,7 +7,6 @@ inherit packagegroup
 SUMMARY_${PN} = "Ostro OS Software updates stack"
 
 RDEPENDS_${PN} = "\
-    tar \
     swupd-client \
     clr-systemd-config \
     "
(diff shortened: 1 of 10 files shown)


kad:
retest this please

pohly:
Yesterday, Mimi Zohar posted a patch ("[Linux-ima-devel] [PATCH 2/2] ima: add support for creating files using the mknodat syscall") which sounds like it might make GNU tar work for us.

Is it worth checking that before switching to bsdtar? The kernel patch looks more intrusive than the one-line change that was needed to make bsdtar work, so I'm not certain whether it applies to our kernel(s) and whether we really want to add it to Ostro.

kad:
I think that having bsdtar for now would be better option, regardless of problem solving between gnu tar and ima. Users should have ability to choose, be bound to GPLv3 code of gnu tar or use different implementation via bsdtar.

pohly:
Right, I had forgotten that we want bsdtar also for other reasons.

rojkov:
I looked at the kernel patch today morning: it doesn't apply to 4.1 cleanly, backporting is needed. Will take a closer look if we are not going to 4.4 any time soon.

barbieri:
Please send a PR to https://github.com/clearlinux/swupd-client and mention me so I can ping Arjan and the team to get these patches in.

rojkov:
@barbieri submitted yesterday - https://github.com/clearlinux/swupd-client/pull/13

Also I submitted a corresponding patch to swud-server.

rojkov:
Added one more commit dropping the dep on tar from swupd packagegroup (https://github.com/ostroproject/meta-ostro/pull/40), rebased and force pushed.

PR #2, opened 2016-02-24
swupd: Disable boot file heuristics

rojkov:
Ostro uses a mechanism different from what's used in ClearLinux
to detect changes in kernel, bootloader, initramfs.
    
This commit disables the CLR specific mechanism by commenting
out a call to boot file heuristics.
    
Fixes: IOTOS-1357

Original PR https://github.com/otcshare/meta-iot-os/pull/242

PR #37, opened 2016-03-10
image-dsk.class: modify rootfs partition size to fit on 4GB media

gvancuts:
It is not uncommon in the IoT space to use 4GB media to host your
OS. 4GB usually means 4.000.000.000 bytes and hence translates
into a real useable size of 3814MB. Setting the size of the rootfs
partition to 3700MB gives us a little buffer and should work with
all 4GB storage media (USB, SD cards, etc.) out there.

### DIFF
--- a/meta-ostro/classes/image-dsk.bbclass
+++ b/meta-ostro/classes/image-dsk.bbclass
@@ -85,7 +85,7 @@ DSK_IMAGE_LAYOUT ??= ' \
     "partition_03_rootfs": { \
         "name": "rootfs", \
         "uuid": "${ROOTFS_PARTUUID_VALUE}", \
-        "size_mb": 5000, \
+        "size_mb": 3700, \
         "source": "${IMAGE_ROOTFS}", \
         "filesystem": "ext4", \
         "type": "8300" \


kad:
:+1:

igor-stoppa:
This is closed but not merged?

igor-stoppa:
+1

@mythi can you please merge this one asap?

mythi:
@kad something is not right with the new settings. I'm getting 

```
This branch is out-of-date with the base branch
Merge the latest changes from master into this branch. 
```

but no change to merge this.

kad:
I fear something wrong going on with status checks on github. I've disabled status checks for now. only force-push protection is enabled now.

kad:
From other hand, this commit is based on top of d7e56f1f0259ba3e02d9da2c8c1f9c22b0fa4752 where master is already 675fcb501b7c1a060405cc1ed42d278d945e1c02. So github might be saying this way that PR require rebase on top of latest master state.

mythi:
@kad thanks

I googled the problem and it seems it's some github problem

kad:
well, still, look how far is base commit for commit in this PR. it might be really request to rebase PR hidden behind such message.

gvancuts:
Do I need to do anything for this? I don't think I do, it looks like it was successfully merged but let me know if that's not the case.

kad:
it was merged without conflicts, but for future PRs we might be more strict about commits being based on very old state of target branch.

mythi:
On Fri, Mar 18, 2016 at 11:12 PM, Alexander D. Kanevskiy wrote:

> it was merged without conflicts, but for future PRs we might be more
> strict about commits being based on very old state of target branch.
>
>
I agree. The question is what is 'very old'. With this one I was sure it
goes OK. Quite often I ask for rebasing because there's a risk
git merge still does OK but combo-layer update in a later phase breaks (git
is able do real with the merge conflict but 'git am' used by
combo-layer fails to apply the patch).

pohly:
On Sun, 2016-03-20 at 23:03 -0700, Mikko Ylinen wrote:
> On Fri, Mar 18, 2016 at 11:12 PM, Alexander D. Kanevskiy wrote:
> 
> > it was merged without conflicts, but for future PRs we might be more
> > strict about commits being based on very old state of target branch.
> >
> >
> I agree. The question is what is 'very old'. With this one I was sure
> it
> goes OK. Quite often I ask for rebasing because there's a risk
> git merge still does OK but combo-layer update in a later phase breaks
> (git
> is able do real with the merge conflict but 'git am' used by
> combo-layer fails to apply the patch).

Isn't that something that our PR testing already detects? I thought the
flow was "check out merge commit in component", then "import that merge
commit via combo-layer". If that works, then I cannot imagine why
"combo-layer" should fail later.

In the past, such problems were not detected because combo-layer was
told to merge the branch head, not the merge commit. This should be
fixed now in the CI scripts.

This detection depends on having up-to-date test results, of course. So
when merging PR #1 in a component, one has to wait for updated test
results for PR #2 before merging the latter.

mythi:
@okartau Can you give an update on this. I remember this was discussed and forgot it when replying. I checked the logs and I cannot see combo-layer is invoked.

kad:
our PR testing scripts are ensuring that combo-layer would work fine if this PR is merged.
However, this error message that GitHub gives is about different thing: if branch protection is active together with status checks, GitHub starts to require that PR content is always rebased on top of latest state of master. So instead of being ok with "git merge --no-ff" it would require that merge of PR would always satisfy FF method of merging to destination branch. I'm not sure that's the best way to do, if there are no conflicts on files, but that's the implementation GitHub has now.

mythi:
@kad OK thanks. @pohly was just correcting my comment ("Quite often I ask for rebasing because there's a risk git merge still does OK but combo-layer update in a later phase breaks").

PR #39, opened 2016-03-11
Doc: Add 802.15.4 setup .rst files

yongli3:
Add the setup documents for 802.15.4 boards

Signed-off-by: Yong Li <yong.b.li@intel.com>


yongli3:
[skip ci]

yongli3:
@dbkinder @gvancuts @lihaofd  @mythi @peijiajames @barbieri @yangbox, I add this sensors support list as .rst file, please help to review

kad:
Maybe it would be better to have Fritzing diagram rather than soldered with single colour of wires picture ? I bet, there are better ways to connect those breakout boards than soldering to galileo pins.

barbieri:
@kad the soldered picture is often helpful, you know, it will match the board and you can check, eventually there are different revisions and the likes and you can easily miss that in a abstract diagram, but it's easier to see in the real picture. But also offering the diagram is good, so we can have a cleaner view of the connections.

doc/hardware/at86rf212.rst
```diff
@@ -0,0 +1,105 @@
...
+
+Reboot the board, the related device driver and kernel modules will be auto loaded, then execute these commands to configure:
+
+On device1:
+
+#. `ip link set wpan0 address a0:0:0:0:0:0:0:1`
+#. `iz set wpan0 777 8001 11`
+#. `ifconfig wpan0 up`
+#. `ip link add link wpan0 name lowpan0 type lowpan`
+#. `ifconfig lowpan0 up`
```
barbieri: err... what happened with the systemd service and scripts you submitted before? They should replace these commands! and those on the other .rst as well

doc/hardware/sensors.rst
```diff
@@ -0,0 +1,41 @@
+
+Sensors list
+##################################
+
+.. table:: Supported sensors on MinnowBoard MAX, Galileo GEN 2 and Edison boards
+
+    ============== ============================================= ==========================================================================
+    Sensor Name     Sensor Type                                    Comment
+    ============== ============================================= ==========================================================================
+    BMA180          Accel                                          missing driver on Edison
```
barbieri: these `missing driver on XXX` won't scale. IMO it's better to use a shorter pattern, we can then have a legend afterwards. Like: `-edison` (= no edison) or `=edison` (= only available in edison)

dbkinder: Does "missing driver..." mean that this sensor is not supported by that platform?  How about changing this to be a column per platform with a "yes" or "no" icon (or text), with blank meaning unknown at this time?   The "yes" entries would then be links (when available) for the underlying document showing how to setup that sensor hardware for that platform (as you've written).

doc/hardware/sensors.rst
```diff
@@ -0,0 +1,41 @@
...
+    LIS3DH          Accel                                          missing driver on Edison
+    MMA7455         Accel                                          missing driver on Edison
+    BMA255          Accel                                          missing driver on Edison
+    LIS2DH12TR      Accel                                          missing driver on Edison
+    BMA222E         Accel                                          missing driver on Edison
+    LSM330DLC       Gyro
+    MPU6050         Gyro
+    MPU9250         Gyro
+    L3G4200         Gyro
+    LSM303          Magnet
```
dbkinder: Since you're sorting this table by "sensor type", a better presentation might be to pull out that column and use it as a section heading and have separate tables for each type.  That would also let you have better sensor type descriptions. For example, "Magnet" should really be "Triple Axis Accelerometer and Magnetometer (Compass)".

dbkinder: You might also link the sensor name (LSM303) to a page that documents that sensor (such as https://learn.adafruit.com/lsm303-accelerometer-slash-compass-breakout/overview).

doc/hardware/cc2520.rst
```diff
@@ -0,0 +1,84 @@
+.. _cc2520:
+
+CC2520 setup
+##################
+
+Category: radio
+Standard: IEEE 802.15.4
+Boards: MinnowBoard MAX
+
+.. image:: _static/cc2520.png
```
dbkinder: I've got some formatting suggestions here, such as adding attributes to control the size and positioning of the image, such as:

```
.. image:: _static/cc2520.png
   :width: 350px
   :alt: CC2520 board 
   :align: right
```

but I can do this kind of layout tweaking after this gets merged and I do an editing pass.

kad:
@barbieri well, even for pictures with actual boards:

- it's better to use colored wires to making connections, as it would be easier for people to trace
- if board has normal pin connectors, it's better to use wires with dupont connectors rather than soldering. it makes better impression usually.

doc/hardware/at86rf212.rst
```diff
@@ -0,0 +1,105 @@
...
+##################
+
+Category: radio
+Standard: IEEE 802.15.4
+Boards: MinnowBoard MAX, Galileo Gen 2
+
+Connection with MinnowBoard MAX board
+
+
+.. image:: _static/at86rf212-minnow.png
```
dbkinder: Wow.  Did you really have to solder and not attach wires to the header pins?

doc/hardware/sensors.rst
barbieri: agreed with all @dbkinder's comments but the column-per-board, as it won't scale when we support more boards :-)

I'd say ideally we create one doc/hardware/XXX.rst, like doc/hardware/lsm303.rst, inside that you can point to adafruit and other places to buy, then also provide a sample usage with FBP and/or JS.

doc/hardware/at86rf212.rst
```diff
@@ -0,0 +1,105 @@
...
+    11    	IRQ
+    12    	MISO
+    13    	DIG4
+    14    	MOSI
+    15		DIG3
+    16		NSEL
+    =====  ==========================
+
+
+.. table:: Connect at86rf212 board to MinnowBoard MAX low speed expansion connector (LSE):
```
dbkinder: While a picture might be nice showing this wiring layout, these tables clearly show what connects to what and work just fine for me.

doc/hardware/at86rf212.rst
yongli3: @dbkinder This radio board does not include soldered pins, by soldering the board to Galileo PCB directly, we can get better electrical signal

doc/hardware/at86rf212.rst
yongli3: Done

yongli3:
@barbieri  in the two radio card .rst files, I replaced these config commands with the systemd service files, please review

barbieri:
ok, that looks good. The per-sensor RST can come in different PR

mythi:
@yongli3 the radio boards setup doc does not match with the latest implementation in https://github.com/ostroproject/meta-ostro-bsp/pull/3 Can you address that please.

I'd also propose to split sensors.rst away from this PR since we have some more opens to it.

yongli3:
@mythi I removed the sensors.rst(will submit another PR later) and modified the .sample file name to match the https://github.com/ostroproject/meta-ostro-bsp/pull/3, please help to review

PR #41, opened 2016-03-11
Revert "ostro_security_flags.inc: Disable -fstack-protector-strong for openjdk"

Valtis:
This reverts commit 9324c2f725ebd382e094ffec26a2a05a2e50615b.

The commit was a temporary fix for a scenario, where builds on certain
compilers were failing. The proper fix for this has been merged, so
this can be reverted.


mythi:
@Valtis Just to make sure: did you test this with your meta-java patch included? I cannot rely on CI builds with this because it was working.

Valtis:
Builds fine on Ubuntu 14.04, which uses gcc 4.8. I rebuilt it with latest master and with this patch just to be sure, and it worked fine.

PR #42, opened 2016-03-11
support xz and zip compressed images

pohly:
This is functionally equivalent to PR #19, but leverages conversion
chaining from OE-core to achieve the same functionality with less
code. This also allows us to use the more natural and user-friendly
dsk.xz (instead of dsk_xz) IMAGE_FSTYPE.

Other benefits:
- all other compression methods also work ("dsk.bz2 dsk.lz4 ...")
- dependencies on the native tools are only added if really
  required
- RM_OLD_IMAGE works for image files

The code can also be made even simpler after enhancing OE-core further.
It is safe to merge this code, too, enhancing OE-core can be done so that
there are no conflicts with the changes in this PR. Both is demonstrated
by https://github.com/pohly/ostro-os/tree/image-compression

### DIFF
--- a/meta-ostro/recipes-image/images/ostro-image.bb
+++ b/meta-ostro/recipes-image/images/ostro-image.bb
@@ -190,6 +190,13 @@ ostro_root_authorized_keys () {
 # Do not create ISO images by default, only HDDIMG will be created (if it gets created at all).
 NOISO = "1"
 
+# Add support for compressing images with zip. Works for arbitrary image
+# types. Example: OSTRO_VM_IMAGE_TYPES = "dsk.zip dsk.vdi.zip"
+COMPRESSIONTYPES_append = " zip"
+ZIP_COMPRESSION_LEVEL ?= "-9"
+COMPRESS_CMD_zip = "zip ${ZIP_COMPRESSION_LEVEL} ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}.zip ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}"
+COMPRESS_DEPENDS_zip = "zip-native"
+
 # Replace the default "live" (aka HDDIMG) images with whole-disk images
 # XXX Drop the VM hack after taking care also of the non UEFI devices (those using U-Boot: edison and beaglebone)
 OSTRO_VM_IMAGE_TYPES ?= "dsk dsk.xz dsk.vdi"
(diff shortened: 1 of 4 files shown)


meta-ostro/recipes-image/images/ostro-image.bb
```diff
@@ -190,9 +190,16 @@ ostro_root_authorized_keys () {
...
+COMPRESSIONTYPES_append = " zip"
+ZIP_COMPRESSION_LEVEL ?= "-9"
+COMPRESS_CMD_zip = "zip ${ZIP_COMPRESSION_LEVEL} ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}.zip ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}"
+COMPRESS_DEPENDS_zip = "zip-native"
+
 # Replace the default "live" (aka HDDIMG) images with whole-disk images
 # XXX Drop the VM hack after taking care also of the non UEFI devices (those using U-Boot: edison and beaglebone)
-OSTRO_VM_IMAGE_TYPES ?= "dsk dsk_xz dsk.vdi"
+OSTRO_VM_IMAGE_TYPES ?= "dsk dsk.xz dsk.vdi"
 IMAGE_FSTYPES_remove_intel-core2-32 = "live"
```
kad: do we really need by default both dsk and dsk.xz ?
I'd say, if we agree that defaults for user would be uncompressed - leave out dsk.xz

pohly: I intentionally kept the behavior of this PR identical to PR #19. But I agree, I think the consensus was to make the default suitable for developers, with .vdi created by default (admittedly that last point is where I lost track of the discussion), so yes, let's change this before merging.

pohly: What I hadn't considered is that disabling compression in this PR would also disable compression in the CI.

Let's deal with changing the default in a separate PR. We need to set up the CI first such that it includes a certain .inc file from meta-ostro, then in that other PR populate the CI settings while changing the non-CI defaults.

@kad: Okay to keep this PR as it is?

kad: ok to keep as is.

igor-stoppa: I know there is a further implementation, which I'll check right away, but my comment on this, so far, is that it still confuses compression and conversion.

mythi:
+1 and tests OK so merging

kad:
side question, can we with such "conversion" mechanism to implement generation of md5/sha sums for images ?

pohly:
On Tue, 2016-03-15 at 05:20 -0700, Alexander D. Kanevskiy wrote:
> side question, can we with such "conversion" mechanism to implement
> generation of md5/sha sums for images ?

Yes, should be doable. However, it'll only work for image files and not
for any of the other published artifacts.

kad:
Image files is the most important part, where checksum is needed :)

PR #43, opened 2016-03-14
To support nfc and add neard to image

wzhen12:
neard is the daemon of linux-nfc.
To support nfc, neard is needed.

Signed-off-by: Wu Zheng <wu.zheng@intel.com>

### DIFF
--- a/meta-ostro/recipes-core/packagegroups/packagegroup-core-connectivity.bb
+++ b/meta-ostro/recipes-core/packagegroups/packagegroup-core-connectivity.bb
@@ -11,4 +11,5 @@ RDEPENDS_${PN} = "\
     connman \
     connman-client \
     lowpan-tools \
+    neard \
     "


wzhen12:
The patch just add "neard" to image.
verify that "neard" has been added to image.

mythi:
test this please

mythi:
@wzhen12 I'd propose this gets added via dependencies. So whoever needs NFC can pull it in.

wzhen12:
@mythi 

You mean that using the systemd dependencies to enable neard?
Such as add "After=neard.service" in someone service file?

Or any suggestion? Thanks.

wzhen12:
@mythi 
 You mean that if anyone want to use neard,  "DEPENDS= neard" need to be added to the related bb files.
It  means that we don't need to do anything for adding neard to image.

The patch will be useless.

If so, it is okay for me.

mythi:
On Wed, Mar 16, 2016 at 9:51 AM, wzhen12 <notifications@github.com> wrote:

> @mythi <https://github.com/mythi>
> You mean that if anyone want to use neard, "DEPENDS= neard" need to be
> added to the related bb files.
> It means that we don't need to do anything for adding neard to image.
>
> The patch will be useless.
>
> If so, it is okay for me.
>
>
Yes, this is what I meant.

mythi:
Mutual agreement this is not needed.

PR #45, opened 2016-03-14
ostro.conf: re-enable python3-* recipes

mythi:
python3-setuptools and python3-pip were PNBLACKLISTed due to errors detected by
oeqa.selftest.iotsstatetests.SStateTests tests for edison vs. intel-core2-32
machines.

python3-setuptools is now needed by the new version of python-gobject recipe so
we need it enabled.

It is possible to just remove the PNBLACKLIST since we have split edison to it's
own package architecture (i.e., not sharing anything with core2-32) AND we are
no longer building with MACHINE=intel-core2-32.


mythi:
fixed a typo and rebased+force pushed.

mythi:
Since it was +1'd by @kad and the builds&tests are good, I'm merging (to the upstream-monitor buildable again)

PR #46, opened 2016-03-14
DOC: Update Edison booting instructions

uhofemei:
Updated flashall.sh location and how to run it
[skip ci]

Signed-off-by: Ulf <ulf.hofemeier@linux.intel.com>

### DIFF
--- a/doc/howtos/booting-and-installation.rst
+++ b/doc/howtos/booting-and-installation.rst
@@ -131,13 +131,12 @@ Flashing an Intel Edison requires use of a breakout board and two micro-USB cabl
    for recovery cases.)
 #. Plug in a micro-USB cable to the J3 connector on the board (corner next to the FTDI chip)
 #. Flip the DIP switch towards jumper J16
-#. Open :command:`minicom` or other terminal program on your host computer to attach to the serial console
-#. Download the ``flashall`` folder from the Ostro OS download folder for edison (on https://download.ostroproject.org)
-#. Copy the flashall script (``flashall.sh``) from the flashall folder to the Ostro OS image folder
-#. Then in the image folder run:: 
-
-    $ sudo ./flashall.sh
-
+#. Download the ``ostro-image`` or ``ostro-image-dev`` image from the Ostro OS download folder for edison (on https://download.ostroproject.org/releases/ostro-os/milestone/)
+#. Extract the image using the command ``$ tar xf ostro-image-*-edison-*.toflash.tar.bz2`` archive
+#. Change directory to the toFlash folder
+   **NOTE:** When flashing the ``-dev`` image please update ``IMAGE_NAME=`` in
+   the flashall.sh script. By default it will look for ``ostro-image``
+#. Run the command ``$ sudo ./flashall.sh``
 #. Plug in the second micro-USB cable to the J16 connector as instructed by the running flashall script
 #. Wait for all the images to flash. You will see the progress on both the flasher and on the serial console
 #. Once flashing is done, the image will automatically boot up and auto-login as ``root``, no password is required


doc/howtos/booting-and-installation.rst
```diff
@@ -131,13 +131,12 @@ Flashing an Intel Edison requires use of a breakout board and two micro-USB cabl
...
-#. Copy the flashall script (``flashall.sh``) from the flashall folder to the Ostro OS image folder
-#. Then in the image folder run:: 
-
-    $ sudo ./flashall.sh
-
+#. Download the ``ostro-image`` or ``ostro-image-dev`` image from the Ostro OS download folder for edison (on https://download.ostroproject.org/releases/ostro-os/milestone/)
+#. Extract the image using the command ``$ tar xf ostro-image-*-edison-*.toflash.tar.bz2`` archive
+#. Change directory to the toFlash folder
+   **NOTE:** When flashing the ``-dev`` image please update ``IMAGE_NAME=`` in
+   the flashall.sh script. By default it will look for ``ostro-image``
```
kad: Instead of modifying script, better to use -i argument to the script.
:-1:

barbieri: this is awful, can't we fix the script we put in the tarball? @jlaako @kad ?

dbkinder: Ah.  When Ulf and I looked at the flashall.sh script, we didn't see -i in the usage description that's displayed.  On closer examination the -i and --image arguments are shown in the example command line that's displayed, but not listed in the usage description text that follows (that should be updated in the script, ja?).

dbkinder: Also, looks like the flashall.bat file (for windows) DOES NOT support the -i option :(

uhofemei:
Fixed.

doc/howtos/booting-and-installation.rst
```diff
@@ -131,13 +131,12 @@ Flashing an Intel Edison requires use of a breakout board and two micro-USB cabl
...
-#. Then in the image folder run:: 
-
-    $ sudo ./flashall.sh
-
+#. Download the ``ostro-image`` or ``ostro-image-dev`` image from the Ostro OS download folder for edison (on https://download.ostroproject.org/releases/ostro-os/milestone/)
+#. Extract the image using the command ``$ tar xf ostro-image-*-edison-*.toflash.tar.bz2`` archive
+#. Change directory to the toFlash folder
+#. Run the command ``$ sudo ./flashall.sh`` |br|
+   **NOTE:** To flash the ``-dev`` image please use the ``-i <image-name>``
+   command line argument
```
barbieri: I still don't like this condition. Either we document all the examples with full command OR we change the included flashall.sh to be smart. My preference is to not need any extra parameter, maybe change the packaging to have a subdir per image and the script can infer the image name from its containing folder?

uhofemei: @barbieri, please remember. Making the script smart means another tool to update if the image naming convention changes. It also means another thing for QA to test. This is why I prefer appending the image name for all images types. It's clean to document and it doesn't require updating the code with naming convention changes.

barbieri: more to user  = more room for errors. If we distribute the tool, then it does make sense to keep them in sync and test. It will also avoid mistakes.

uhofemei: Your decision. For now we need the documentation update merged as the instructions how to flash Edison on the Ostro website don't work. I will take care of updating the docs once the flashall script has been enhanced.

uhofemei:
@mythi, I agree and this merge request adds documentation that works.

mythi:
@uhofemei in fact, after double-checking the images and the previous PRs, what @barbieri is asking is already implemented: each ```.toflash.``` image has the IMAGE_NAME correctly set so no need to add parameters/edit anything.

uhofemei:
@mythi  that makes sense then. I ran into it because I used the M3 image for Edison we are pointing to on the ostroproject.org website.

kad:
Script was fixed quite a long time ago in ostroproject/meta-ostro-bsp#4. However, this would not fix situation with old published build 399. So, it's safe to refer users to use -i argument, even if default value would be already good enough. Actually, @jlaako in ostroproject/meta-ostro-bsp#5 removed default value. So if user builds image by himself, he would anyway need to use -i argument.

Bad part, that Linux flasher script is not on par with Windows bat script in that functionality.

uhofemei:
I updated the Edison flashing instructions.

kad:
@uhofemei  can you please squash your changes into one commit and rebase it on top of master ? merge commits are not really good thing in PRs.

PR #48, opened 2016-03-15
ostro.conf: Set libmicrohttpd preferred version

brunobottazzini:
To build Soletta with http server support the libmicrohttpd
should be the 0.9.43 or newer.

### DIFF
--- a/meta-ostro/conf/distro/ostro.conf
+++ b/meta-ostro/conf/distro/ostro.conf
@@ -212,6 +212,9 @@ PNWHITELIST_meta-soletta += " \
     graphviz \
 "
 
+# Soletta requires minimum 0.9.43 to build it with http server support
+PREFERRED_VERSION_libmicrohttpd ?= "0.9.43"
+
 # Java bootstrapping
 PREFERRED_PROVIDER_virtual/java-initial-native = "cacao-initial-native"
 PREFERRED_PROVIDER_virtual/java-native = "jamvm-native"


mythi:
@brunobottazzini while you are at it, perhaps add systemd to DEPENDS as well (it's not related to this PR but something I had noticed earlier)

brunobottazzini:
@mythi, AFAIK Soletta does not depends on systemd, with exception the nodetype platform/service

I will add this depends with "bb.utils(distro_feature, systemd)" to make sure that the nodetype will be built when systemd is enabled.

barbieri:
soletta itself will depend on systemd if platform=systemd, it's not just node types. 

Ideally you'd detect systemd in distro_features and force/check for the .config to contain that platform, otherwise force micro-linux platform.

mythi:
On Tue, Mar 15, 2016 at 4:00 PM, Gustavo Sverzut Barbieri  wrote:

> soletta itself will depend on systemd if platform=systemd, it's not just
> node types.
>
In meta-ostro we currently have a soletta bbappend that adds DEPENDS_append
= " systemd"  because of
"platform=systemd" in our .config. AFAICS, Kconfig defaults to this since
HAVE_SYSTEMD gets detected.

barbieri:
indeed, it will get detected.

what I mean is to play well and have a .bb that supports non-systemd systems, we can do not depend on it and force platform=micro-linux. Otherwise systemd will get detected yet we'll not depend on it.

kad:
so, is this PR just about pre-caution that other layers would be providing different version of libmicrohttpd ? or is it intended to be particularly need of 0.9.43 or later ? if later, it would be better to have in soletta-dev-app_git.bb: "libmicrohttpd (>= 0.9.43)".

brunobottazzini:
@kad, Yes there is other recipe of libmicrohttpd in meta-openembedded and we'd like to use the 0.9.43 that is included in meta-soletta. 

This patch will make sure that we will use the right one and Soletta will be built with http-server support.

This dependency is from Soletta and not the Soletta Dev-App.

kad:
@brunobottazzini name of bb recipe I've got from git grep in meta-soletta directory.
If you need it in soletta, then you probably missing dependency there.

brunobottazzini:
@kad I have added the dependency there when I was testing. It wasn't enough, though. 

I had to define the preferred version, or remove the openembedded version that I think it is not a good think to do.

kad:
have you added just dependency to libmicrohttpd or versioned (>= ...) ?

brunobottazzini:
@kad, I have tested like this: https://github.com/solettaproject/meta-soletta/pull/48

I was unaware about the existence of  ">= 0.9.43". 

I shall test with DEPENDS="libmicrohttpd (>= 0.9.43)" and tell you if it will work

thank you very much

mythi:
Actually, hold on. Things are clear now.

we have libmicrohttpd whitelisted from meta-oe. I had missed the version dependency in soletta. So I created 4682aeaa.

The right fix is to drop 4682aeaa and whitelist the lib from meta-soletta (and set the right DEPENDS)

brunobottazzini:
@mythi, thanks! I will test with your patch.

In the message commit replace Soletta's Dev-App to Soletta. 

It will enable Soletta to have http-server module to be built with it.

mythi:
@brunobottazzini the commit is pretty old (and already merged). 

The patch you need to add here is

```
--- a/meta-ostro/conf/distro/ostro.conf
+++ b/meta-ostro/conf/distro/ostro.conf
@@ -148,7 +148,6 @@ PNWHITELIST_networking-layer += " \
 PNWHITELIST_openembedded-layer += " \
     gmock \
     gtest \
-    libmicrohttpd \
     libsocketcan \
     can-utils \
     giflib \
@@ -205,6 +204,7 @@ PNWHITELIST_meta-java += " \
 PNWHITELIST_meta-soletta += " \
     soletta \
     soletta-dev-app \
+    libmicrohttpd \
     lss \
     python3-jsonschema \
     python3-jsonschema-native \
```

brunobottazzini:
This PR can be closed since this change was addressed here: https://github.com/ostroproject/meta-ostro/pull/62

PR #50, opened 2016-03-15
Set defaults for both local development and CI

igor-stoppa:
An individual developer is probably not interested in producing
compressed images.

Instead, it is more likely to use images that are deployable either
on the target or in a Virtual Machine, therefore do not waste
time/space compressing them by default.

This can be configured in:

meta-ostro/conf/distro/include/ostro-os-development.inc
meta-ostro/conf/distro/include/ostro-os-production.inc

The Continuous Integration infrastructure, otoh, wants to
produce only images that are convenient to download.

This can be configured in:

meta-ostro/conf/distro/include/ostroproject-ci.inc

Signed-off-by: Igor Stoppa <igor.stoppa@intel.com>

### DIFF
--- a/meta-ostro/conf/distro/include/ostroproject-ci.inc
+++ b/meta-ostro/conf/distro/include/ostroproject-ci.inc
@@ -1,0 +1,5 @@
+# XXX Drop the VM hack after taking care also of the non UEFI devices (those using U-Boot: edison and beaglebone)
+OSTRO_VM_IMAGE_TYPES = "dsk.zip dsk.vdi.zip"
+
+# Include the devel settings
+require conf/distro/include/ostro-os-development.inc
--- a/meta-ostro/recipes-image/images/ostro-image.bb
+++ b/meta-ostro/recipes-image/images/ostro-image.bb
@@ -182,7 +182,7 @@ COMPRESS_DEPENDS_zip = "zip-native"
 
 # Replace the default "live" (aka HDDIMG) images with whole-disk images
 # XXX Drop the VM hack after taking care also of the non UEFI devices (those using U-Boot: edison and beaglebone)
-OSTRO_VM_IMAGE_TYPES ?= "dsk dsk.xz dsk.vdi"
+OSTRO_VM_IMAGE_TYPES ?= "dsk dsk.vdi"
 IMAGE_FSTYPES_remove_intel-core2-32 = "live"
 IMAGE_FSTYPES_append_intel-core2-32 = " ${OSTRO_VM_IMAGE_TYPES}"
 IMAGE_FSTYPES_remove_intel-corei7-64 = "live"


igor-stoppa:
This should be merged only when @kad and @okartau confirm that they are ready with CI.

meta-ostro/conf/distro/include/ostroproject-ci.inc
```diff
@@ -0,0 +1,24 @@
+# Use the pre-generated keys for IMA signing.
+IMA_EVM_KEY_DIR = "${IMA_EVM_BASE}/data/debug-keys"
+
```
kad: those should not be here.

meta-ostro/conf/distro/include/ostroproject-ci.inc
```diff
@@ -0,0 +1,24 @@
...
+# Enable local root access.
+OSTRO_EXTRA_IMAGE_FEATURES += "autologin"
+
```
kad: this is also to ostro-os-development.inc.

meta-ostro/conf/distro/include/ostroproject-ci.inc
```diff
@@ -0,0 +1,24 @@
...
+OSTRO_EXTRA_MOTD () {
+************************************
+*** This is a development image! ***
+*** Do not use in production.    ***
+************************************
+}
+
+# Everything ready, avoid initial sanity check.
+OSTRO_IMAGE_BUILD_MODE_SELECTED = "1"
+
```
kad: this should not be here as well.

meta-ostro/conf/distro/include/ostro-os-development.inc
```diff
@@ -15,3 +15,16 @@ OSTRO_EXTRA_MOTD () {
...
+# contain the string "VM" even when it refers to real machines.
+
+# Image for UEFI devices.
+# OSTRO_VM_IMAGE_TYPES = " dsk"
+
+# Image for VirtualBox.
+OSTRO_VM_IMAGE_TYPES = " dsk.vdi"
+
+# Both images.
+# OSTRO_VM_IMAGE_TYPES = " dsk dsk.vdi"
```
kad: those should be "?=" assignments, so if user have different order of includes in his local.conf, it would be not overwriting user's choice.

pohly: The purpose of ostro-os-development.inc and ostro-os-production.inc was to switch between production and developer images, i.e. these files control the content of the images. The format of the images is orthogonal to that and thus should not be changed by these files.

My expectation was that ostroproject-ci.inc add those config changes that need to be set differently in the CI compared to what is the default when building straight after "oe-init-build-env". One such change is "require ostro-os-development.inc", the other is enabling the compressed image formats.

In addition, the default value of OSTRO_VM_IMAGE_TYPES in ostro-image.bb needs to be changed such that compression is not listed.

igor-stoppa: On 15 March 2016 at 21:06, Patrick Ohly <notifications@github.com> wrote:
>
> The purpose of ostro-os-development.inc and ostro-os-production.inc was to
> switch between production and developer images,
>

ok, ​I had a different interpretation, but I can change the patches to this
one.



-- 
igor

mythi:
@igor-stoppa are you working on an update to this?

igor-stoppa:
On 17 March 2016 at 17:15, Mikko Ylinen <notifications@github.com> wrote:

> @igor-stoppa <https://github.com/igor-stoppa> are you working on an
> update to this?
>
>
> ​yes and I even have it finished, I got side-tracked by the swupd
integration

I'll brew a tea and submit the new version of this one​



-- 
igor

igor-stoppa:
@mythi apologies, after the tea came Sasha and the rest of the day took an unforeseen turn.
Here's the updated patchset.

meta-ostro/conf/distro/include/ostroproject-ci.inc
```diff
@@ -0,0 +1,2 @@
+# XXX Drop the VM hack after taking care also of the non UEFI devices (those using U-Boot: edison and beaglebone)
+OSTRO_VM_IMAGE_TYPES = "dski.zip dsk.vdi.zip"
```
pohly: Typo: s/dski.zip/dsk.zip/

To address kad's earlier concerns for the previous patches, overriding OSTRO_VM_IMAGE_TYPES unconditionally here is fine, because these are exactly the values the CI is to use, i.e. nothing needs to change them.

@kad How do you want the CI to use include files? Include both "ostro-development.inc" and "ostroproject-ci.inc", or just "ostroproject-ci.inc"? I had the impression that it is the latter, in which case this file here must include ostro-development.inc.

igor-stoppa: On 18 March 2016 at 11:23, Patrick Ohly <notifications@github.com> wrote:

> In meta-ostro/conf/distro/include/ostroproject-ci.inc
> <https://github.com/ostroproject/meta-ostro/pull/50#discussion_r56629858>:
>
> > @@ -0,0 +1,2 @@
> > +# XXX Drop the VM hack after taking care also of the non UEFI devices (those using U-Boot: edison and beaglebone)
> > +OSTRO_VM_IMAGE_TYPES = "dski.zip dsk.vdi.zip"
>
> Typo: s/dski.zip/dsk.zip/
>

​fixed, thanks.​


> @kad <https://github.com/kad> How do you want the CI to use include
> files? Include both "ostro-development.inc" and "ostroproject-ci.inc", or
> just "ostroproject-ci.inc"? I had the impression that it is the latter, in
> which case this file here must include
> ​​
> ostro-development.inc.
>
>
​Doesn't CI need to produce also images based on ostro-os-production.inc?​


​Wouldn't the inclusion here of ​​ostro-development.inc here prevent those?




-- 
igor

kad:
CI would include ostroproject-ci.inc, which would control everything else.
We are not producing "ostro-os-production" related images.

igor-stoppa:
On 18 March 2016 at 12:34, Alexander D. Kanevskiy <notifications@github.com>
wrote:

> CI would include ostroproject-ci.inc, which would control everything else.
> We are not producing "ostro-os-production" related images
> ​.
>
>
​Ok, please have a look at the updated patch.

-- 
igor

kad:
Looks good to me. :+1:

meta-ostro/conf/distro/include/ostroproject-ci.inc
pohly: On Fri, 2016-03-18 at 03:19 -0700, Igor Stoppa wrote:


> ​Doesn't CI need to produce also images based on
> ostro-os-production.inc?​ ​Wouldn't the inclusion here of ​​
> ostro-development.inc here prevent those?

It is impossible to produce images for both production and development
in the same build configuration; the CI exclusively configures itself to
build development images in all jobs.

If we want to change that (I believe we should, at least to ensure that
a production configuration builds, but that's a different story), we'd
have to add a "ostroproject-production-ci.inc" that gets included in
different build jobs and re-arrange our .inc files accordingly.

pohly:
Yes, okay to be merged.

mythi:
+1 from me too. The CI scripts fail since the .xz files are now missing. I'll sync with @okartau before merging.

kad:
retest this please

kad:
Temporary compression added to the scripts, so let's re-test. In previous run, edison was still failing

mythi:
obsolete.

PR #53, opened 2016-03-15
DOC: update flashing instructions for Edison board

dbkinder:
Updated flashall.sh location and how to run it
  - flashall.sh script is now in the *.toflash.tar.bz2 tarball
  - post build 399 (M3), the script is specific to the image downloaded (ostro-image and ostro-image-dev)
    but with build 399 it isn't so a "-i <imagename>" argument was added to let the user override the
    default imagename (keep that documented for now)

[skip ci]

### DIFF
--- a/doc/howtos/booting-and-installation.rst
+++ b/doc/howtos/booting-and-installation.rst
@@ -4,7 +4,7 @@ Booting and Installing an Ostro |trade| OS Image
 #################################################
 
 This technical note explains the basic procedures for taking an Ostro OS image that was downloaded
-or built from source (using instructions in :ref:`Building Images`), and installing and 
+or built from source (using instructions in :ref:`Building Images`), and installing and
 running that image one of the :ref:`platforms`.
 
 Two images are of interest for this process (depending if you're using real hardware or a VM):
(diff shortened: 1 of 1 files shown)


dbkinder:
Sorry about the removing trailing whitespace edits included here.  The significant changes are lines 128-150 for the Edison instructions.  This PR replaces PR#26

uhofemei:
:+1:

doc/howtos/booting-and-installation.rst
```diff
@@ -127,47 +128,52 @@ Intel Edison
...
+
+   $ tar xf ostro-image-*-edison-*.toflash.tar.bz2
+
+#. Change directory to the toFlash folder.
+#. Run the command::
+
+   $ sudo ./flashall.sh
 
-    $ sudo ./flashall.sh
+   `NOTE:` If the script is unable to find the image, use the ``-i <imagename>`` option to the flashall script.
```
mythi: Note: with #399 this *is* the case if user downloads ostro-image-dev and flashes it.

mythi:
@dbkinder can you address the feedback

dbkinder:
I think that statement does address the problem, no?  If you get an error (as you will with 399) then try again using the -i <imagename> option.  What else should we say?

mythi:
@dbkinder Indeed, I was thinking #399 case only (in which it is likely user needs two attempts :)

Looks good.

PR #54, opened 2016-03-16
systemd: wait for udev settle in systemd-networkd

mythi:
In order to make USB networking start reliably at boot time, wait
udev to settle before launching systemd-network. Makes results of
network-online.target consistent.

### DIFF
--- a/meta-ostro/recipes-core/systemd/systemd/0001-systemd-networkd.service-wait-for-udev-to-settle.patch
+++ b/meta-ostro/recipes-core/systemd/systemd/0001-systemd-networkd.service-wait-for-udev-to-settle.patch
@@ -1,0 +1,30 @@
+From ecc712dca90b3c1a86d033d16a590fd794d7d92c Mon Sep 17 00:00:00 2001
+From: Jussi Laako <jussi.laako@linux.intel.com>
+Date: Mon, 14 Mar 2016 16:07:32 +0200
+Subject: [PATCH] systemd-networkd.service: wait for udev to settle
+
+In order to make USB networking start reliably at boot time, wait udev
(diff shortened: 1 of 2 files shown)


jlaako:
Looks good, needs bunch of testing to make sure it really solves the issue. Kind of hard to test one, because the issue it is fixing appears randomly...

mythi:
The first attempt still gives a failure: https://ostroproject.org/jenkins/job/test_edison/202/

test this please

kad:
I suspect something else got broken... in serial log it shows that image booted up to console and sshd already started, but test cases are not able to reconnect to device.

mythi:
@kad I found a pattern in the serial.log files. Something triggers fsck and that timeouts.

@jlaako @Valtis

kad:
unclean shutdown ? RTC (time goes back) ?

jlaako:
fsck at shutdown time? Or next bootup?

Even if there is unclean shutdown, there's no point in running fsck with ext4...

If there is unclean shutdown it is indication that filesystem unmount failed and that's usually because some process is left running that keeps filesystem busy...

mythi:
@jlaako it seems already in first boot

See e.g., https://ostroproject.org/jenkins/job/test_edison/240/artifact/serial.log

jlaako:
@mythi That log looks like a complete filesystem corruption and failure...

I have a feeling that I was seeing something possibly related on MMax yesterday, where systemd decided to trigger reboot during first boot, possibly due to fsck, although it didn't say why it decided to reboot (and no fsck output on serial console).

mythi:
We now seem to have a good fix in meta-ostro-bsp so closing this one.

PR #56, opened 2016-03-16
ostro-image.bb: always install app-framework by default.

klihub:
One of the conculsions in the recent QA-related discussions was
that QA wants to be able to run some app-framework tests using
an unmodified image. To enable this, added a dedicated package
group for app-framework and added that to the list of default
package groups to install. This will ensure that the necessary
bits are in the image even without any applications present.

The old combined app-fw/security-fw/user-management package
group does not make sense and probably should be removed. It
is left intact in place for now.

### DIFF
--- a/meta-ostro/recipes-image/images/ostro-image.bb
+++ b/meta-ostro/recipes-image/images/ostro-image.bb
@@ -32,6 +32,7 @@ IMAGE_FEATURES[validitems] += " \
     soletta-tools \
     swupd \
     tools-develop \
+    app-framework \
 "
 
 # These features come from base recipes, but are not added to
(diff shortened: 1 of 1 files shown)


mythi:
Needs rebasing too.

pohly:
On Wed, 2016-03-16 at 07:46 -0700, Krisztian Litkey wrote:
> The old combined app-fw/security-fw/user-management package
> group does not make sense and probably should be removed. It
> is left intact in place for now.

If it does not make sense anymore, then please remove it, either in this
PR or a separate one.

mythi:
Let's drop app-privileges in this PR.

mythi:
@klihub I'm closing this since QA is now happy with the example-app-* (that also pull in appfw) being installed in the -dev image.

please reopen this first if/once you make changes. however, IMO, this could wait until the bundle stuff is finished.

PR #61, opened 2016-03-17
local.conf.sample: remove Qemu configuration block

gvancuts:
Remove the Qemu configuration block as QEMU is not supported
with Ostro due to its EFI implementation limitations

### DIFF
--- a/meta-ostro/conf/layer.conf
+++ b/meta-ostro/conf/layer.conf
@@ -17,7 +17,7 @@ BBFILE_PRIORITY_ostro = "6"
 # need to be increased by one each time a change is made to
 # local.conf.sample that requires manually updating a local.conf after
 # updating the meta data.
-LOCALCONF_VERSION = "1"
+LOCALCONF_VERSION = "2"
 
 # Same for LCONF_VERSION in bblayer.conf.sample.
 LAYER_CONF_VERSION = "7"
--- a/meta-ostro/conf/local.conf.sample
+++ b/meta-ostro/conf/local.conf.sample
@@ -232,22 +232,10 @@ BB_DISKMON_DIRS = "\
 # Example for Ostro OS setup, recommended to use it:
 #SSTATE_MIRRORS ?= "file://.* http://download.ostroproject.org/sstate/ostro-os/PATH"
 
-
-#
-# Qemu configuration
-#
-# By default qemu will build with a builtin VNC server where graphical output can be
-# seen. The two lines below enable the SDL backend too. This assumes there is a
-# libsdl library available on your build system.
-PACKAGECONFIG_append_pn-qemu-native = " sdl"
-PACKAGECONFIG_append_pn-nativesdk-qemu = " sdl"
-ASSUME_PROVIDED += "libsdl-native"
-
-
 # CONF_VERSION is increased each time build/conf/ changes incompatibly and is used to
 # track the version of this file when it was generated. This can safely be ignored if
 # this doesn't mean anything to you.
-CONF_VERSION = "1"
+CONF_VERSION = "2"
 
 #
 # ISAFW: https://github.com/01org/isafw


kad:
we need to check, how that is done currently in Yocto. they might already deprecated that usage of SDL from host.

gvancuts:
Assuming we don't use Qemu at all anymore in Ostro, whether or not they still use SDL is somewhat irrelevant to us, isn't it?

kad:
qemu is built for image conversion, and probably some hidden dependencies in OE-core regardless we use it or not. SDL bits might be crucial, depending on how it's build in different distros.

gvancuts:
I had forgotten about the image conversion thing. But thinking about it, isn't that qemu-native rather that Qemu?

I've noticed that one check failed but that seems related to an SSH reconnection issue. I'm not too sure what to do about it, any idea?

kad:
it's always qemu-native as I recall for all use cases. Well, looking at this patch, it might be also nativesdk-qemu. 

For check failure - yes, this looks like overall edison instability in test lab :(

mythi:
@gvancuts this justifies CONF_VERSION change. Can make that change?

In #75 (and in some of the oe-core changes that are queued) @pohly suggests changes that make the error notes clear enough for users.

mythi:
```
# This and the corresponding CONF_VERSION value in local.conf.sample
# need to be increased by one each time a change is made to
# local.conf.sample that requires manually updating a local.conf after
# updating the meta data.
```

(in meta-ostro/conf/layer.conf)

gvancuts:
I've bumped up CONF_VERSION in meta-ostro/conf/local.conf.sample. Is this what you were asking for @mythi ?

I'm a little confused by your reference to meta-ostro/conf/layer.conf... does that one *also* need to be bumped up?

mythi:
@gvancuts yes, both (as written in the comment)

pohly:
On Mon, 2016-03-21 at 07:00 -0700, Geoffroy Van Cutsem wrote:
> I've bumped up CONF_VERSION in meta-ostro/conf/local.conf.sample. Is
> this what you were asking for @mythi ?

Also bump LOCALCONF_VERSION in meta-ostro/conf/layer.conf. When someone
copies the updated local.conf.sample, the versions match and nothing
happens.

But when later we bump LOCALCONF_VERSION again and someone uses it with
a local.conf that still has the previous version, the sanity check kicks
in and warns the user.

gvancuts:
OK, I find this mechanism odd (or at least I don't fully understand its intent and how it works)... but anyway, I'll force-push anew patch in a minute.

gvancuts:
Force-pushed an updated patch. Thanks

mythi:
@gvancuts thanks. LOCALCONF_VERSION is the version we check user has in his local.conf. If we make changes to our local.conf.sample (CONF_VERSION changes), we need to update LOCALCONF_VERSION to match the new version we are providing.

gvancuts:
Thanks @mythi , why don't (or can't) we check CONF_VERSION directly? If someone updates the sources, check the CONF_VERSION in meta-ostro/conf/local.conf.sample against the one found in ${BUILD}/conf/local.conf and update it if necessary?

mythi:
You need variables with different names.

gvancuts:
That could be hidden in the code that does the check, couldn't it? Compare the same variable from two different locations or even assign them different names but in your code, not something that is visible to the user. Asking someone to keep two different variables in sync is asking for trouble.

pohly:
On Mon, 2016-03-21 at 09:12 -0700, Geoffroy Van Cutsem wrote:
> Thanks @mythi , why don't (or can't) we check CONF_VERSION directly?
> If someone updates the sources, check the CONF_VERSION in
> meta-ostro/conf/local.conf.sample against the one found in
> ${BUILD}/conf/local.conf and update it if necessary?

I guess it is more complicated than that because "bitbake" at runtime no
longer knows where the local.conf.sample came from (that's something
that oe-init-build-env figures out), so the value also has to be set in
a file that gets parsed by bitbake.

gvancuts:
Could oe-init-build-env store such variable? I mean something like:
- source oe-init-build-env -> sets SAMPLE_CONF_VERSION (or any other name) to what's in local.conf.sample
- When you invoke bitbake, it derives the CONF_VERSION variable from your local.conf file and compares it with SAMPLE_CONF_VERSION

pohly:
On Mon, 2016-03-21 at 10:09 -0700, Geoffroy Van Cutsem wrote:
> Could oe-init-build-env store such variable? I mean something like:
> 
>       * source oe-init-build-env -> sets SAMPLE_CONF_VERSION (or any
>         other name) to what's in local.conf.sample

Set where? In the local conf directory, for example in local.conf?

>       * When you invoke bitbake, it derives the CONF_VERSION variable
>         from your local.conf file and compares it with
>         SAMPLE_CONF_VERSION

That wouldn't work, because in such a scenario, the SAMPLE_CONF_VERSION
in local.conf and the CONF_VERSION in local.conf would continue to match
even when the repo gets updated to a newer version.

What would work is to somehow store a link back to the original
local.conf.sample. But quite frankly, I find that a pretty complex and
therefore fragile approach (what if the local.conf.sample location
changes?) that just isn't worth implementing.

mythi:
On Mon, Mar 21, 2016 at 6:30 PM, Geoffroy Van Cutsem  wrote:

> That could be hidden in the code that does the check, couldn't it? Compare
> the same variable from two different locations or even assign them
> different names but in your code, not something that is visible to the
> user. Asking someone to keep two different variables in sync is asking for
> trouble.
>

btw, this isn't something users would see. It only involves us as
meta-ostro local.conf.sample provides/maintainers.

gvancuts:
@pohly , I was more thinking of an env variable that would simply exist once you've source the oe-init-build-env script, not something that would be set at runtime by bitbake.

But not to worry, it's probably more hassle than what we'd gain from it.

mythi:
@gvancuts can you rebase this on top of latest master (there are merge conflicts after I merged #75)

gvancuts:
Just force-pushed a rebased patch.

mythi:
test this please

mythi:
test this please

mythi:
merging despite fully unclean build. the changes do not affect testing.

PR #70, opened 2016-03-18
Enable BMAP (block map) Tools usage in Ostro

kad:
- Ability to generate image checksums using custom compression/conversion mechanisms.
- Ability to generate BMAP files for images.

### DIFF
--- a/doc/howtos/booting-and-installation.rst
+++ b/doc/howtos/booting-and-installation.rst
@@ -27,19 +27,77 @@ additional build and debugging tools that wouldn't typically be included in a pr
 dev image also will auto-login as ``root`` at the console, something that normally would not be available
 in a production device image but is quite useful during development.
 
-Using dd to Create Bootable Media
-=================================
+
+Using bmaptool to Create Bootable Media
+=======================================
 
 Once you have the :file:`.dsk.xz` Ostro OS image you need to get it
 onto your hardware platform, typically by using removable media such as a
-USB thumb drive or SD card.  The usual way to do this is with the :command:`dd` command.
+USB thumb drive or SD card.
+
+The recommended way to do this is with the :command:`bmaptool` command from `bmap-tools`_.
+A copy of this utility is available in the :file:`deploy/tools` folder after a Yocto Project build
+for your image is finished.
+
+The ``bmaptool`` program automatically handles copying either compressed and uncompressed images to
+your removable media.  It also also uses a generated ``image.bmap`` file containing a checksum for
+itself and for all mapped data regions in the image file, making it possible to verify data integrity
+of the downloaded image. Be sure to download this ``.bmap`` file along with the image for your device.
+
+
+#. Connect your USB thumb drive or SD card to your Linux-based development system
+   (minimum 4 GB card required, for some images 8 GB card might be required).
+#. If you're not sure about your media device name, use the :command:`dmesg` command to view the system log
+   and see which device the USB thumb drive or SD card was assigned (e.g. :file:`/dev/sdb`)::
+
+      $ dmesg
+
+   or you can use the :command:`lsblk` command to show the block-level devices; a USB drive usually
+   shows up as ``/sdb`` or ``/sdc``
+   (almost never as ``/sda``) and an SD card would usually show up as :file:`/dev/mmcblk0`.
+
+   Note: You should specify the whole device you're writing to with
+   :command:`bmaptool`:  (e.g., :file:`/dev/sdb` or
+   :file:`/dev/mmcblk0`) and **not** just a partition on that device (e.g., :file:`/dev/sdb1` or
+   :file:`/dev/mmcblk0p1`) on that device.
+
+#. The :command:`bmaptool` command will overwrite all content on the device so be careful specifying
+   the correct media device. The ``bmaptool`` opens the removable media exclusively and helps prevent
+   writing on an unintended device. After verifying your removable media device name, you'll need
+   to ``umount`` the device before writing to it.
+
+   In the example below, :file:`/dev/sdb` is the
+   destination USB device on our development machine::
+
+      $ sudo umount /dev/sdb*
+      $ sudo -E bmaptool copy <ostro-os-image> /dev/sdb
+
+.. note::
+    The :command:`bmaptool` is intelligent enough to recognize images in different
+    formats, including compressed images (.gz, .bz2, .xz) as well as flashing
+    directly from remote URL (for example, you could specify the image source file with an
+    ``http://`` address instead of downloading it first; ``bmaptool`` will automatically retrieve
+    the .bmap file). The ``sudo -E`` option will propagate environment variables (such as http_proxy)
+    that bmaptool might need to access the website.
+
+
+Unplug the removable media from your development system and you're ready to plug
+it into your target system.
+
+.. _bmap-tools: http://git.infradead.org/users/dedekind/bmap-tools.git/blob/HEAD:/docs/README
+
+Using dd to Create Bootable Media
+=================================
+
+While using ``bmaptool``  to create your bootable media is preferred because it's faster and
+includes a checksum verification, you can also use the traditional :command:`dd` command instead :
 
 #. Connect your USB thumb drive or SD card to your Linux-based development system
    (minimum 8 GB card required).
 #. If you're not sure about your media device name, use the :command:`dmesg` command to view the system log
    and see which device the USB thumb drive or SD card was assigned (e.g. :file:`/dev/sdb`)::
 
-     $ dmesg
+      $ dmesg
 
    or you can use the :command:`lsblk` command to show the block-level devices; a USB drive usually
    shows up as ``/sdb`` or ``/sdc``
@@ -106,7 +164,7 @@ MinnowBoard MAX from the USB thumbdrive described above.
 Galileo Gen 2
 =============
 
-The `Intel Galileo Gen 2`_ is an Intel® Quark x1000 32-bit, single core, Intel Pentium |reg| Processor class 
+The `Intel Galileo Gen 2`_ is an Intel® Quark x1000 32-bit, single core, Intel Pentium |reg| Processor class
 SOC-based board, pin-compatible with shields designed for the Arduino Uno R3.
 
 Flashing an `Intel Galileo Gen 2`_ requires use of a microSD card (booting off USB is not supported).
@@ -147,7 +205,7 @@ Flashing an Intel Edison requires use of a breakout board and two micro-USB cabl
 #. Plug in the second micro-USB cable to the J16 connector as instructed by the running flashall script.
 #. Wait for all the images to flash. You will see the progress on the flasher.
 #. Once flashing is done, the image will automatically boot up and auto-login as ``root``, no password is required.
-   
+
 BeagleBone Black
 ================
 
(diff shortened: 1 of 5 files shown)


gvancuts:
Thanks for doing this @kad !! This looks good to me although I'm no expert in classes, etc.

@dbkinder , we should be ready to add this to our documentation when it gets merged. We should specifically document that there will be checksum files available by default and we should also document how to use the bmap-tool.

meta-ostro/classes/ostro-image-extensions.bbclass
```diff
@@ -0,0 +1,21 @@
...
+COMPRESS_CMD_md5sum = "md5sum ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type} > ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}.md5sum"
+COMPRESS_CMD_sha1sum = "sha1sum ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type} > ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}.sha1sum"
+COMPRESS_CMD_sha224sum = "sha224sum ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type} > ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}.sha224sum"
+COMPRESS_CMD_sha256sum = "sha256sum ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type} > ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}.sha256sum"
+COMPRESS_CMD_sha384sum = "sha384sum ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type} > ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}.sha384sum"
+COMPRESS_CMD_sha512sum = "sha512sum ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type} > ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}.sha512sum"
+
+COMPRESS_DEPENDS_vdi = "qemu-native"
+COMPRESS_DEPENDS_bmap = "bmap-tools-native"
+COMPRESS_DEPENDS_md5sum = "coreutils-native"
```
pohly: We should check with the Yocto team whether we really have to force the compilation of coreutils-native or can assume that this is present.

kad: Ideally, this should be converted to use python implementation of hashlib functions, but I'm not sure how to specify that in bbclass syntax. if you have idea how to do that, please share.

As for coreutils-native, we definetely want to get @rpurdie, @rossburton feedback, but I'd prefer to play safe. if somebody wants to skip build of it, it's possible to set "assume provided" in local.conf for it.

kad:
@igor-stoppa @mythi please have a look at this idea.
@jlaako - we need to find way to use same mechanisms for files generated for Edison and BeagleBone.

@pohly you were faster to notice PR than I expected :) I wanted to ask you later, after I see first automated results, to have a look for two items (suggest PR to my forked repo to image-extensions branch, so we can accumulate changes):
- check how we can adjust build of bmap-tools, so it can be used to actually flash the image locally using bmap-tool.
- see how we can ship bmaptool inside SDK/eSDK that user can use it to flash images. Maybe we would need to have some help here from @bluelightning

@gvancuts we indeed need to document that, but first I'd like to figure out how to ship bmap tool in such way that it can be used by user not only for generation of bmap, but also for flashing without installing it to host OS. For that we need to solve above items.

@igor-stoppa one thing I've noticed by looking to result bmap file, and afterwards in code for generating dsk images: we are using for some file operations "dd", which looses sparsness of the files. result is dsk image 99.8% mapped, vdi 100% mapped, where in reality they are not. We need to replace calls to external dd and truncate to python native code, so generation of combined image will not loose sparsness of the the result file.

meta-ostro/recipes-image/images/ostro-image.bb
```diff
@@ -182,7 +182,7 @@ COMPRESS_DEPENDS_zip = "zip-native"
 
 # Replace the default "live" (aka HDDIMG) images with whole-disk images
 # XXX Drop the VM hack after taking care also of the non UEFI devices (those using U-Boot: edison and beaglebone)
-OSTRO_VM_IMAGE_TYPES ?= "dsk dsk.xz dsk.vdi"
+OSTRO_VM_IMAGE_TYPES ?= "dsk dsk.sha256sum dsk.bmap dsk.xz dsk.vdi dsk.vdi.sha256sum dsk.vdi.bmap"
```
pohly: Does it make sense to provide bmap data for .vdi images? These images are not going to get copied to any install media, are they?

pohly:
On Sat, 2016-03-19 at 01:03 -0700, Alexander D. Kanevskiy wrote:
> @pohly you were faster to notice PR than I expected :) I wanted to ask
> you later, after I see first automated results, to have a look for two
> items (suggest PR to my forked repo to image-extensions branch, so we
> can accumulate changes):
> 
>       * check how we can adjust build of bmap-tools, so it can be used
>         to actually flash the image locally using bmap-tool.
>       * see how we can ship bmaptool inside SDK/eSDK that user can use
>         it to flash images. Maybe we would need to have some help here
>         from @bluelightning

Sorry, I don't have any bandwidth left to tackle new problems right now.
But for whoever gets to work on this, can you clarify what you mean with
"flash image locally"? Inside a running Ostro OS, or on some arbitrary
Linux host?

meta-ostro/recipes-image/images/ostro-image.bb
kad: potentially, you can use bmaptool to efficiently copy files from remote location with preserving sparseness on filesystem.

kad:
@pohly  what I mean by "flash image locally", is to get something like
```
. oe-init-build-env
bitbake ostro-image
bitbake bmap-tools-native
sudo bmaptool tmp-glibc/deploy/images/intel-corei7-64/ostro-image-intel-corei7-64-*.dsk /dev/sdc
```

At the moment, last step is not easily possible, without installing bmap-tools to host OS.
I would like to have a way that this tool is usable from bitbake generated sysroot rather than installed to host OS.

meta-ostro/recipes-image/images/ostro-image.bb
pohly: On Mon, 2016-03-21 at 02:51 -0700, Alexander D. Kanevskiy wrote:
> In meta-ostro/recipes-image/images/ostro-image.bb:
> 
> > @@ -182,7 +182,7 @@ COMPRESS_DEPENDS_zip = "zip-native"
> >  
> >  # Replace the default "live" (aka HDDIMG) images with whole-disk images
> >  # XXX Drop the VM hack after taking care also of the non UEFI devices (those using U-Boot: edison and beaglebone)
> > -OSTRO_VM_IMAGE_TYPES ?= "dsk dsk.xz dsk.vdi"
> > +OSTRO_VM_IMAGE_TYPES ?= "dsk dsk.sha256sum dsk.bmap dsk.xz dsk.vdi dsk.vdi.sha256sum dsk.vdi.bmap"
> 
> potentially, you can use bmaptool to efficiently copy files from
> remote location with preserving sparseness on filesystem.

I thought .vdi files do not have much sparseness already? IMHO that's
worth testing before adding more complexity here and in the
corresponding user documentation.

kad: In my tests they some times have about 1% of sparseness, but that's not important, or potentially can be because of issues with loosing sparseness on dsk image creation time. Anyway, setting here is just example at the moment, to validate that it works correctly in automated build. Actual settings would be adjusted later via ostroproject-ci.inc.

jlaako:
@kad this kind of stuff should go to openembedded/yocto. We don't have bandwidth to maintain any more complex BSP customizations than we currently have.

kad:
@jlaako those customization are working fine for any image that is produced normal yocto way. However, as I can see in edison case, tasks that are generating images (tarballs) are overriden in meta-ostro-bsp, thus where those definition would be in either OE-core or in meta-ostro layer, this wouldn't make much difference as Edison image generation doesn't use them.

meta-ostro/classes/ostro-image-extensions.bbclass
rpurdie: I guess the best way of handling this would be to put coreutils-native into ASSUME_PROVIDED. I think assuming its present is a pretty good assumption as we certainly use md5sum elsewhere.

kad:
@rpurdie can you please validate commit 22e4178 if I'm using it correctly to put standalon version of tool to deploy/tools ?

kad:
test this please

kad:
test this please

kad:
@dbkinder Draft of documentation included in 18abc5d commit. 
Please help me here with improving it, based on available docs from past projects. e.g. https://source.tizen.org/documentation/reference/bmaptool or http://git.infradead.org/users/dedekind/bmap-tools.git/blob/HEAD:/docs/README

kad:
@dbkinder thanks for review. proposed changed incorporated.

doc/howtos/booting-and-installation.rst
```diff
@@ -27,19 +27,79 @@ additional build and debugging tools that wouldn't typically be included in a pr
...
+A copy of this utility is available in the :file:`deploy/tools` folder after a Yocto Project build
+for your image is finished, or you can install the bmap-tools package with your development system's
+package manager.  For example, on Ubuntu/Debian systems use ::
+
+      $ sudo apt-get install bmap-tools
+
+The ``bmaptool`` program automatically handles copying either compressed and uncompressed images to
+your removable media.  It also also uses a generated ``image.bmap`` file containing a checksum for
+itself and for all mapped data regions in the image file, making it possible to verify data integrity
+of the downloaded image. Be sure to download this ``.bmap`` file along with the image for your device.
```
igor-stoppa: Shouldn't it work also with urls?

doc/howtos/booting-and-installation.rst
```diff
@@ -27,19 +27,79 @@ additional build and debugging tools that wouldn't typically be included in a pr
...
+   (almost never as ``/sda``) and an SD card would usually show up as :file:`/dev/mmcblk0`.
+
+   Note: You should specify the whole device you're writing to with
+   :command:`bmaptool`:  (e.g., :file:`/dev/sdb` or
+   :file:`/dev/mmcblk0`) and **not** just a partition on that device (e.g., :file:`/dev/sdb1` or
+   :file:`/dev/mmcblk0p1`) on that device.
+
+#. The :command:`bmaptool` command will overwrite all content on the device so be careful specifying
+   the correct media device. The ``bmaptool`` opens the removable media exclusively, helping prevent
+   writing to an unintended device.  After verifying your removable media device name, you'll need
```
igor-stoppa: @dbkinder I take the chance to nitpick, mostly out of personal curiosity.
Is it correct to say "helping prevent writing"? To my, admittedly, non-native English-speaker ear, this sounds unnatural.

meta-ostro/classes/ostro-image-extensions.bbclass
```diff
@@ -0,0 +1,15 @@
+# Class that provides additional tools and methods for image conversions.
```
igor-stoppa: I would drop "ostro" from the name of the class. There isn't really anything specific to ostro.

meta-ostro/lib/image-dsk.py
```diff
@@ -1,56 +1,105 @@
...
+
 def do_dsk_image():
+    """Entry point for generating the disk image."""
     # Load the descripton of the disk layout.
     partition_table = json.loads(expand_vars('${DSK_IMAGE_LAYOUT}'))
 
     # Before adding up the size of each partition, add the size of the GPT
     full_image_size_mb = partition_table["gpt_initial_offset_mb"] + \
-                         partition_table["gpt_tail_padding_mb"]
+        partition_table["gpt_tail_padding_mb"]
```
igor-stoppa: nitpicking, iirc this sort of indentation is superseded by the one being replaced

doc/howtos/booting-and-installation.rst
kad: Actually, this sentence about "download bmap" can be avoided. Tool can handle URLs and can automatically detect location of bmap file if present, even over remote like HTTP/HTTPS.

meta-ostro/classes/ostro-image-extensions.bbclass
kad: patch for those submitted upstream, but not sure how long it would take to accept it.
My idea was that after functionality verification, I'll move bmap releated things into separate class where image-dsk would be inheriting from. vdi would stay in image-dsk, and all checksum related things would go to upstream. meanwhile, it can be in ostro distro configs (not even needed to be in bbclass to be available globally).

meta-ostro/lib/image-dsk.py
kad: that's my bad habit of having tools that verify and auto-reformat code according to PEP8 in my text editor :)

doc/howtos/booting-and-installation.rst
dbkinder: It is a bit awkward now that I read it again... let's try "... opens the removable media exclusively and helps prevent writing on an unintended device."

doc/howtos/booting-and-installation.rst
dbkinder: bmaptool does say it works with urls, but if you naively go to the release download page, there's a link to the compressed image file for you to download.  If you just grab that and try to use bmaptool you'll get an error since it can't find the bmap file.

kad: That's the case, because bmap indeed not present on download site.
In case of such cases it's always possible to use '--nobmap' flag.
Might be way to improve fallback mechanism in future.

doc/howtos/booting-and-installation.rst
kad: Updated.

kad:
Removed all development flags. Now this branch contains cleaned version of BMAP support (code+documentation), so it can be merged.

mythi:
@kad please rebase this one more time, thanks

mythi:
@kad also, your oe-core patch is now merged (ea11c46)

kad:
Haven't noticed that my patch was merged recently. (as usual, no visibility on patch flow in MLs). 
Cleaned a bit, rebased and pushed.

mythi:
+1 

fails to build because of sha256 IMAGE_CMD missing but that is queued in ostro-os and coming from oe-core update so we can merge this.

PR #71, opened 2016-03-19
DOC: Add HowTo for configuring IP address with ConnMan

dbkinder:
howtos/ip-address-config.rst
  describe how to use ConnMan to configure a static or dynanic IP addresses
  for a WiFi or Ethernet connection

howtos/howtos.rst
  added ip-address-config to list of technical notes

[skip ci]

### DIFF
--- a/doc/howtos/ip-address-config.rst
+++ b/doc/howtos/ip-address-config.rst
@@ -1,0 +1,119 @@
+.. _ip-address-config:
+
+Configuring an IP Address in the Ostro |trade| OS
+######################################################
+
+The Ostro Project uses Connection Manager (ConnMan), a Linux daemon and command-line utility, for managing internet connections 
+within embedded devices.
+
+ConnMan supports these network technologies:
+
+* Ethernet, using a built-in card, connected via host or client side USB
+* WiFi, using wpa_supplicant
+* Celluar connectivity using oFono
+* Bluetooth connectivity using BlueZ
+
+By default all technologies are disabled at the very first startup
+to prevent unwanted wireless or wired communication from happening. ConnMan remembers the state of network
+technologies over reboots. 
+(You can read more about ConnMan at https://01.org/connman/documentation.)
+
+This tech note explains how to connect your device to a network manually using the ``conmanctl`` command-line client.  For proper integration in 
+an application, you should use the D-Bus APIs (and their bindings).  
+
+
+Enabling Network Technologies 
+=============================
+
+Before a network transport (such as WiFi) can be used, it must be enabled in ConnMan. You can check to see what technologies
+have already been enabled with the command :: 
+
+   $ connmanctl technologies
+
+The command will indicate if LAN and WiFi technologies are enabled. If not, you can enable either (or both) of them with the commands:: 
+
+   $ connmanctl enable ethernet 
+   $ connmanctl enable wifi 
+
+
+ConnMan will automatically handle wired connections with a DHCP server.
+
+Configuring a Static IP Address
+================================
+
+ConnMan refers to network devices as services. To configure a static IP address for a given LAN or WiFi service, 
+you first find the name of the service assigned by ConnMan and then provide the IP address information, as 
+shown in these steps: 
+
+#. List the services available on your device ::
+
+     $ connmanctl services
+
+     *AO Wired                ethernet_0008a209b525_cable
+         AnOpenNetwork        wifi_dc85de828967_4d6568657272696e_managed_none
+
+   The wired service name discovered is ``ethernet_0008a209b525_cable``.
+
+   Note: The symbols in the output above are: '\*' favorite (saved) network, 'A' autoconnectable, 'O' online and 'R' ready. 
+   If no letter is shown in the O/R column, the network is not connected. In addition, temporary states include 
+   'a' for association, 'c' configuration and 'd' disconnecting. When any of these three letters are showing, 
+   the network is in the process of connecting or disconnecting. A network is in state 'ready' once it has 
+   obtained an IPv4 or IPv6 address or both, and can indicate that the network might still need a proxy set up to get 
+   internet connectivity.
+
+#. Once you know the service name, you can assign a static IP with the command::
+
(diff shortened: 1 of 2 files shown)


doc/howtos/ip-address-config.rst
```diff
@@ -0,0 +1,103 @@
...
+
+The Ostro Project uses Connection Manager (ConnMan), a Linux daemon and command-line utility, for managing internet connections 
+within embedded devices. 
+
+ConnMan supports these network technologies:
+
+* Ethernet, using a built-in card, connected via host or client side USB
+* WiFi, using wpa_supplicant
+* Celluar connectivity using oFono
+* Bluetooth connectivity using Bluez
```
gvancuts: It's spelled BlueZ (capital Z)

doc/howtos/ip-address-config.rst
```diff
@@ -0,0 +1,103 @@
...
+     $ connmanctl config <service> –ipv4 manual <ip address> <netmask> <gateway> 
+
+   For example, using the service name discovered above: ::
+
+     $ connmanctl config ethernet_0008a209b525_cable –ipv4 manual 192.168.1.4 255.255.255.0 192.168.1.1
+
+You can assign a static IP to a discovered WiFi adapter service in the same manner.
+
+
+Configuring a Dynamic (DHCP) IP Address
```
gvancuts: I think it'd be more appropriate to call this section something like "Connecting to a WiFi network".

gvancuts: Or add a note in the paragraph below on how to use DHCP with Ethernet (it's basically the default behaviour so enabling ethernet is all that's needed)

uhofemei: I can look into this more on Monday when I get back into the office. I don't know what the default configuration of Connman in Ostro looks like. The default behavior of Connman upstream is to have all technologies disabled by default for security purposes.

gvancuts:
:+1:

doc/howtos/ip-address-config.rst
```diff
@@ -0,0 +1,103 @@
+.. _ip-address-config:
+
+Configuring an IP Address in the Ostro |trade| OS
+######################################################
+
+The Ostro Project uses Connection Manager (ConnMan), a Linux daemon and command-line utility, for managing internet connections 
+within embedded devices.
```
barbieri: Although correct I'd rephrase to put some emphasis on the DBus API and that it is official way to talk to connman. Then you can follow using connmanctl, then we add some DBus call that does the same, people can then use Python DBus or something like that in real products instead of trying to call system() on connmanctl

Soletta and nodejs will have nice C/FBP/JS version, so we can refer to them later.

doc/howtos/ip-address-config.rst
```diff
@@ -0,0 +1,103 @@
...
+Here are the steps to connect to a wireless network using DHCP:
+
+#. Open connmanctl in interactive mode, enable WiFi (if you haven't already), and scan for new WiFi services (available SSIDs)::
+
+     $ connmanctl
+     $ connmanctl> enable wifi 
+     $ connmanctl> scan wifi
+     $ connmanctl> services 
+
+#. Take note of the wireless network name and the service name displayed by the ``services`` command.  To connect to a secure
```
barbieri: Agent is very important concept, add more emphasis and explain it is used by the daemon to call back an application when attention or input is needed. In the command line tool, available in development images, once enabled it will work until you quit the interactive session. In real life products the application is expected to always keep it alive.

doc/howtos/ip-address-config.rst
dbkinder: I think you're saying that there's a programmatic way to configure IP addresses, and while true, the immediate need was to document how to do this manually.  I'm sure there are plenty of code examples (like this) we need to write about explaining how to do things, that would be appropriate for an application programmer's cookbook with complete examples.

gvancuts: I'd suggest we add a line to explain that this documents explains how to connect your device to a network 'manually' and _just_ mention that for proper integration in an app, people should look at the D-Bus API (and its bindings). Would that work for both of you?

dbkinder:
force pushed updates to address Gustavo and Geoffroy's comments.  Programmatic (d-bus API) ip address configuration out of scope for this document, but should be covered in an application developer's guide/cookbook (TBD)

doc/howtos/ip-address-config.rst
```diff
@@ -0,0 +1,119 @@
...
+* WiFi, using wpa_supplicant
+* Celluar connectivity using oFono
+* Bluetooth connectivity using BlueZ
+
+By default all technologies are disabled at the very first startup
+to prevent unwanted wireless or wired communication from happening. ConnMan remembers the state of network
+technologies over reboots. 
+(You can read more about ConnMan at https://01.org/connman/documentation.)
+
+This tech note explains how to connect your device to a network manually using the ``conmannctl`` command line client.  For proper integration in
```
mythi: conmannctl -> connmanctl

doc/howtos/ip-address-config.rst
```diff
@@ -0,0 +1,119 @@
...
+* Celluar connectivity using oFono
+* Bluetooth connectivity using BlueZ
+
+By default all technologies are disabled at the very first startup
+to prevent unwanted wireless or wired communication from happening. ConnMan remembers the state of network
+technologies over reboots. 
+(You can read more about ConnMan at https://01.org/connman/documentation.)
+
+This tech note explains how to connect your device to a network manually using the ``conmannctl`` command line client.  For proper integration in 
+an application, the D-Bus service API (and its bindings) should be used.
```
gvancuts: A couple of nitpicks:
- Shouldn't command line be spelled with an hyphen, i.e. command-line?
- Technically speaking, we'd expect people to use more than just ConnMan service API for proper integration, they'd also use its Manager API, Session API, etc. so I'd suggest we just say "... the D-Bus API (and its bindings) should be used."

barbieri:
fair points, current document looks good to me. Later when we add connman in soletta and node.js we can link to those documents from here.

dbkinder:
forced push to address comments

PR #72, opened 2016-03-20
ostro.conf: enable uninative

kad:
Enable usage of uninative feature for native components.
This would allow to re-use same native components across
different Linux distributions.
Config file comes from OE-Core, thus we can be sure that
it always points to corresponding to our state of OE-Core
shim glibc version, and this version already tested and
used within Yocto Project.

### DIFF
--- a/meta-ostro/conf/distro/ostro.conf
+++ b/meta-ostro/conf/distro/ostro.conf
@@ -306,3 +306,7 @@ INPUTANALYZER_WHITELIST = '/(meta|meta-yocto-bsp|meta-intel|meta-java|meta-oic|m
 # MASK incompatible .bbappends from meta-soletta. meta-soletta's systemd
 # .bbappend is in conflict with our needs.
 BBMASK = "meta-soletta/recipes-configure/systemd"
+
+# re-use uninative shim released by Yocto Project / OE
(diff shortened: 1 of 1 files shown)


pohly:
The configuration file points towards http://downloads.yoctoproject.org/releases/uninative/0.95/. Enabling the feature will cause bitbake to download some components for the native SDK from the Yocto server. Will it use the Ostro OS content delivery network as pre-mirror or at least as fallback?

kad:
it will be present in our mirror directory (fetcher puts it to same directory as any other downloaded sources), so our settings PREMIRRORS would be serving it from our servers, but primary location is yoctoproject.org. We are not going to maintain our own copy or release it separately from OE.

pohly:
On Sun, 2016-03-20 at 13:31 -0700, Alexander D. Kanevskiy wrote:
> it will be present in our mirror directory (fetcher puts it to same
> directory as any other downloaded sources), so our settings PREMIRRORS
> would be serving it from our servers, but primary location is
> yoctoproject.org. We are not going to maintain our own copy or release
> it separately from OE.

That's what I was hoping for :-)

+1 from me for merging.

mythi:
+1 too! will merge

PR #76, opened 2016-03-23
ostro-image-dev: add dev-pkgs in IMAGE_FEATURES

mythi:
For a development image that installs all development tools, including
the compilers, it is also useful to have the -dev packages for all
components installed.

To get that, add "dev-pkgs" image feature in IMAGE_FEATURES for the
-dev variant.


gvancuts:
:+1:

mythi:
-dev packages will be handled via -dev bundles so this is not needed.

PR #82, opened 2016-03-24
Virtualbox appliance generation

Valtis:
Add virtualbox recipe and virtualbox appliance generation as additional image type. Virtualbox appliances are currently generated for intel-corei7-64 and intel-quark images.

This is, at this stage, primarily meant to trigger CI testing and gather feedback. In no particular order:

* Recipes for virtualbox and libidl (which is dependency for virtualbox) are currently placed under meta-ostro/recipes-core. I'm not sure if this is the correct place.
* The shell function that handles the appliance generation is in image-dsk.bbclass. I'm not sure if this is proper place for this either.
* There was discussion on mailing list, that using intel-quark as the 32-bit variant is likely a bad idea, as it is highly machine dependent. Perhaps machine independent 32 bit image should be used instead (intel-corei7-32?)
* libidl recipe was originally in meta-layer, but was removed at some point. Brief googling did not reveal if it was completely removed, or moved to some other layer. The recipe is from here: http://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/packages/libidl/libidl_0.8.12.bb?id=8917b6eb2999f8a894c4864246ed34ef56ff00b9
 
### DIFF
--- a/meta-ostro/classes/image-dsk.bbclass
+++ b/meta-ostro/classes/image-dsk.bbclass
@@ -25,11 +25,12 @@
 
 
 # Ostro custom conversion types
-COMPRESSIONTYPES_append = " vdi bmap"
+COMPRESSIONTYPES_append = " vdi bmap ova"
 COMPRESS_CMD_vdi = "qemu-img convert -O vdi ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type} ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}.vdi"
 COMPRESS_CMD_bmap = "bmaptool create ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type} -o ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}.bmap"
 COMPRESS_DEPENDS_vdi = "qemu-native"
 COMPRESS_DEPENDS_bmap = "bmap-tools-native"
+COMPRESS_DEPENDS_ova = "vboxmanage-native"
 
 # Needed to use native python libraries
 inherit pythonnative
@@ -37,6 +38,78 @@ inherit pythonnative
 # Image files of machines using image-dsk.bbclass do not use the redundant ".rootfs"
 # suffix. Probably should be moved to ostro-os.conf eventually.
 IMAGE_NAME_SUFFIX = ""
+
+create_ova() {
+
+  if [ "x${MACHINE}x" != "xintel-corei7-64x" ]; then
+    exit 0
+  fi
+
+  # 64 bit linux with sata bus
+  OS_TYPE="Linux_64"
+  STORAGE_NAME="SATA"
+  STORAGE_BUS_TYPE="sata"
+
+  VM_NAME="${IMAGE_NAME}"
+  # 512 should be large enough value to run Ostro (Galileo has 256MB RAM for example)
+  # but small enough value that even a modest host machine can afford it
+  RAM="512"
+  # min allowed - as we do not run graphical environment, it should be enough
+  VRAM="12"
+  # As above with RAM, 2 cores should be enough for Ostro without crippling the host machine
+  CPU_CORE_COUNT=2
+  # required for successful boot
+  FIRMWARE="efi"
+  IOAPIC="on"
+
+  # default settings for hard drive
+  STORAGE_PORT="0"
+  STORAGE_DEVICE="0"
+  STORAGE_DEVICE_TYPE="hdd"
+
+  
+  RAW_IMAGE="${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.dsk"
+  VIRTUALBOX_IMAGE="${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.vmdk"
+  VIRTUALBOX_EXECUTABLE="${STAGING_BINDIR_NATIVE}/VBoxManage"
+
+  APPLIANCE_NAME="${VM_NAME}.dsk.ova"
+
+  FOLDER_NAME="virtualbox-vm"
+  WORK_FOLDER="${DEPLOY_DIR_IMAGE}/${FOLDER_NAME}"
+
+  mkdir -p ${WORK_FOLDER}
+
+  # erase any temp files from previous builds, as these can cause issues
+  # notably NS_ERROR_FACTORY_NOT_REGISTERED (0x80040154)
+  rm -rf /tmp/.vbox-root-ipc
+
+  # create vm
+  ${VIRTUALBOX_EXECUTABLE} createvm --name ${VM_NAME} --ostype ${OS_TYPE} --basefolder ${WORK_FOLDER} --register
+
+  #set core count, memory, firmware and ioapic
+  ${VIRTUALBOX_EXECUTABLE} modifyvm ${VM_NAME} --memory ${RAM} --vram ${VRAM} --cpus ${CPU_CORE_COUNT} --firmware ${FIRMWARE} --ioapic ${IOAPIC}
+
+  # add bus for hard drives
+  ${VIRTUALBOX_EXECUTABLE} storagectl ${VM_NAME} --name ${STORAGE_NAME} --add ${STORAGE_BUS_TYPE}
+  
+  # create virtual hard drive from the raw .dsk image
+  ${VIRTUALBOX_EXECUTABLE} internalcommands createrawvmdk -filename ${VIRTUALBOX_IMAGE} -rawdisk ${RAW_IMAGE}
+  
+  # attach the .vmdk image as hard drive
+  ${VIRTUALBOX_EXECUTABLE} storageattach ${VM_NAME} --storagectl ${STORAGE_NAME} --medium ${VIRTUALBOX_IMAGE} --port ${STORAGE_PORT} --device ${STORAGE_DEVICE} --type ${STORAGE_DEVICE_TYPE}
+
+  # export the image
+  ${VIRTUALBOX_EXECUTABLE} export ${VM_NAME} --output ${DEPLOY_DIR_IMAGE}/${APPLIANCE_NAME} --ovf20
+
+  # unregister the VM from virtualbox - 
+  ${VIRTUALBOX_EXECUTABLE} unregistervm  ${VM_NAME}
+
+  # chmod the outputs so that they can be accessed by non-owners as well
+  chmod +r ${APPLIANCE_NAME}
+  chmod +r ${VIRTUALBOX_IMAGE}
+}
+
+COMPRESS_CMD_ova = "create_ova"
 
 do_uefiapp[depends] += " \
                          systemd:do_deploy \
--- a/meta-ostro/recipes-core/vboxmanage/vboxmanage-native_5.0.16.bb
+++ b/meta-ostro/recipes-core/vboxmanage/vboxmanage-native_5.0.16.bb
@@ -1,0 +1,88 @@
+LICENSE="GPLv2"
+LIC_FILES_CHKSUM="file://COPYING;md5=e197d5641bb35b29d46ca8c4bf7f2660"
+
+inherit native
+
+BPN="VirtualBox"
+
+DEPENDS=" \
+    libxml2-native glib-2.0-native libidl-native acpica-native cdrtools-native \
+    libpng-native zlib-native openssl-native curl-native libcap-native \
+    makeself-native libpam-native libxslt-native \
+"
+
+SRC_URI=" \
+    http://download.virtualbox.org/virtualbox/${PV}/VirtualBox-${PV}.tar.bz2 \
+    file://configure.patch \
+    file://xpidl-build-fixes.patch \
+    file://include-dir.patch \
+    file://remove-x11.patch \
+    file://fix-linker-flags.patch \
+    file://disable-guest-additions.patch \
+"
+
+SRC_URI[md5sum] = "1752a485b1cb377cee5f196918cda741"
+SRC_URI[sha256sum] = "f5a44d33a1db911f445b2eb2d22d9293a96a535cba488b5a235577ef868fa63c"
+
+# export path to native libdir, as this is needed during build
+# (see xpidl build fix patch)
+export VIRTUALBOX_YOCTO_LIBDIR="${STAGING_LIBDIR_NATIVE}"
+
+# The build files seem to override LDFLAGS, so re-export it with different name and patch files to use it
+export VIRTUALBOX_YOCTO_LDFLAGS="${LDFLAGS}"
+
+# Disable everything we do not need (which turns out to be almost everything)
+VBOX_CONF = " \
+    --disable-hardening \
+    --disable-python \
+    --disable-java \
+    --disable-vmmraw \
+    --disable-sdl-ttf \
+    --disable-alsa \
+    --disable-pulse \
+    --disable-dbus \
+    --disable-kmods \
+    --disable-opengl \
+    --disable-extpack \
+    --disable-docs \
+    --disable-libvpx \
+    --disable-udptunnel \
+    --disable-devmapper \
+    --disable-vmmraw \
+    --build-headless \
+    --with-makeself='${STAGING_BINDIR_NATIVE}/makeself.sh' \
+    "
+
+# As we do not use either autotools or cmake, we have to call configure manually
+do_configure() {
+    ./configure ${VBOX_CONF}
+}
+
+do_compile() {
+    . ./env.sh
+    kmk
+}
+
+do_install() {
+    # minimal set of files for VBoxManage
+    ARCH_DIR=`ls ${B}/out`
+
+    install -d ${D}${libdir}/vboxmanage
+    install -m 0755 ${B}/out/${ARCH_DIR}/release/bin/VBoxManage ${D}${libdir}/vboxmanage
+    install -m 0644 ${B}/out/${ARCH_DIR}/release/bin/VBoxDDU.so ${D}${libdir}/vboxmanage
+    install -m 0644 ${B}/out/${ARCH_DIR}/release/bin/VBoxRT.so ${D}${libdir}/vboxmanage
+    install -m 0644 ${B}/out/${ARCH_DIR}/release/bin/VBoxXPCOM.so ${D}${libdir}/vboxmanage
+    install -m 0755 ${B}/out/${ARCH_DIR}/release/bin/VBoxSVC ${D}${libdir}/vboxmanage
+    install -m 0755 ${B}/out/${ARCH_DIR}/release/bin/VBoxXPCOMIPCD ${D}${libdir}/vboxmanage
+
+    install -d ${D}${libdir}/vboxmanage/components
+    install -m 0644 ${B}/out/${ARCH_DIR}/release/bin/components/VBoxC.so ${D}${libdir}/vboxmanage/components
+    install -m 0644 ${B}/out/${ARCH_DIR}/release/bin/components/VBoxSVCM.so ${D}${libdir}/vboxmanage/components
+    install -m 0644 ${B}/out/${ARCH_DIR}/release/bin/components/VBoxXPCOMBase.xpt ${D}${libdir}/vboxmanage/components
+    install -m 0644 ${B}/out/${ARCH_DIR}/release/bin/components/VBoxXPCOMIPCC.so ${D}${libdir}/vboxmanage/components
+    install -m 0644 ${B}/out/${ARCH_DIR}/release/bin/components/VirtualBox_XPCOM.xpt ${D}${libdir}/vboxmanage/components
+
+
+    install -d ${D}${bindir}
+    ln -sf ../lib/vboxmanage/VBoxManage ${D}${bindir}/VBoxManage
+}
--- a/meta-ostro/classes/ostro-image.bbclass
+++ b/meta-ostro/classes/ostro-image.bbclass
@@ -238,7 +238,7 @@ COMPRESS_DEPENDS_zip = "zip-native"
 
 # Replace the default "live" (aka HDDIMG) images with whole-disk images
 # XXX Drop the VM hack after taking care also of the non UEFI devices (those using U-Boot: edison and beaglebone)
-OSTRO_VM_IMAGE_TYPES ?= "dsk dsk.vdi"
+OSTRO_VM_IMAGE_TYPES ?= "dsk dsk.vdi dsk.ova"
 IMAGE_FSTYPES_remove_intel-core2-32 = "live"
 IMAGE_FSTYPES_append_intel-core2-32 = " ${OSTRO_VM_IMAGE_TYPES}"
 IMAGE_FSTYPES_remove_intel-corei7-64 = "live"
--- a/meta-ostro/conf/bblayers.conf.sample
+++ b/meta-ostro/conf/bblayers.conf.sample
@@ -34,7 +34,7 @@ OSTRO_LAYERS = " \
 
 # OSTRO_LAYERS += "##OEROOT##/meta-openembedded/meta-efl"
 # OSTRO_LAYERS += "##OEROOT##/meta-openembedded/meta-filesystems"
-# OSTRO_LAYERS += "##OEROOT##/meta-openembedded/meta-gnome"
+OSTRO_LAYERS += "##OEROOT##/meta-openembedded/meta-gnome"
 # OSTRO_LAYERS += "##OEROOT##/meta-openembedded/meta-gpe"
 # OSTRO_LAYERS += "##OEROOT##/meta-openembedded/meta-initramfs"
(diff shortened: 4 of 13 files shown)


meta-ostro/recipes-core/libidl/libidl_0.8.14.bb
```diff
@@ -0,0 +1,22 @@
+LICENSE = "LGPL"
+DESCRIPTION = "Library for parsing CORBA IDL files"
```
mythi: let's whitelist this from meta-gnome

meta-ostro/recipes-core/virtualbox/virtualbox.bb
```diff
@@ -0,0 +1,75 @@
...
+
+# export path to native libdir, as this is needed during build
+# (see xpidl build fix patch)
+export VIRTUALBOX_YOCTO_LIBDIR="${STAGING_LIBDIR_NATIVE}"
+
+SRC_URI[md5sum] = "1752a485b1cb377cee5f196918cda741"
+SRC_URI[sha256sum] = "f5a44d33a1db911f445b2eb2d22d9293a96a535cba488b5a235577ef868fa63c"
+
+# Disable everything we do not need (which turns out to be almost everything)
+CONFIGS = " \
```
mythi: Use ```EXTRA_OECONF```

meta-ostro/recipes-core/virtualbox/virtualbox.bb
```diff
@@ -0,0 +1,75 @@
...
+    install -m 0644 ${B}/out/${ARCH_DIR}/release/bin/components/VirtualBox_XPCOM.xpt ${D}${libdir}/virtualbox/components
+
+
+    install -d ${D}${bindir}
+    ln -sf ../lib/virtualbox/VBoxManage ${D}${bindir}/VBoxManage
+
+
+}
+
+BBCLASSEXTEND += "native"
```
mythi: I'd explicitly make the recipe virtualbox-native.bb

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -24,16 +24,104 @@
...
 python () {
     if d.getVar('IMAGE_DSK_ACTIVE', True) == 'True':
         d.setVar('IMAGE_NAME_SUFFIX', '')
 }
 
+
+create_ova() {
+
+
+  if [ "x${MACHINE}x" = "xintel-corei7-64x" -o "x${MACHINE}x" = "xintel-quarkx" ]; then
```
mythi: Let's start with intel-corei7-64 only. The if [] can exit if not intel-corei7-64.

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -24,16 +24,104 @@
...
+    # attach the .vdi image as hard drive
+    ${VIRTUALBOX_EXECUTABLE} storageattach ${VM_NAME} --storagectl ${STORAGE_NAME} --medium ${VIRTUALBOX_IMAGE} --port ${STORAGE_PORT} --device ${STORAGE_DEVICE} --type ${STORAGE_DEVICE_TYPE}
+
+    # Ensure we are in correct working directory
+    cd ${DEPLOY_DIR_IMAGE}
+
+    ${VIRTUALBOX_EXECUTABLE} export ${VM_NAME} --output ${APPLIANCE_NAME} --ovf20
+
+    # delete old symlink
+    rm -f ${SYMLINK_NAME}
```
mythi: the symlink handling should come from image.bbclass, no?

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -24,16 +24,104 @@
...
+    elif [ "x${MACHINE}x" = "xintel-quarkx" ]; then
+      OS_TYPE="Linux"
+      STORAGE_NAME="IDE"
+      STORAGE_BUS_TYPE="ide"
+    fi
+
+    VM_NAME="${IMAGE_NAME}"
+    RAM="512"
+    VRAM="12"
+    CPU_CORE_COUNT=2
```
mythi: It'd be good to add a comment how the defaults got selected and that they can be modified.

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -24,16 +24,104 @@
...
+    SYMLINK_NAME="${BPN}-${MACHINE}.ova"
+
+    FOLDER_NAME="virtualbox_vm"
+    WORK_FOLDER="${DEPLOY_DIR_IMAGE}/${FOLDER_NAME}"
+
+    mkdir -p ${WORK_FOLDER}
+
+    # erase any temp files from previous builds, as these can cause issues
+    # notably NS_ERROR_FACTORY_NOT_REGISTERED (0x80040154)
+    cd /tmp/
```
mythi: Is ```VIRTUALBOX_EXECUTABLE``` using both /tmp and ```WORK_FOLDER```?

mythi:
@igor-stoppa can you take a look as well and provide your feedback?

Valtis:
Addressed some of the feedback:

* libidl is now whitelisted from meta-gnome (I didn't realize it was present already)
* Renamed virtualbox recipe to vboxmanage-native, to better highlight it that it is not a full virtualbox, but rather a minimal subset to enable the VBoxManage tool
* vboxmanage recipe uses EXTRA_OECONF for configuration parameters
* Changed appliance creation shell function only to create virtualbox appliance for 64 bit intel i7 image, and added some comments to explain why certain values are chosen. 

Symlink creation was not addressed, as I need to look into it a bit more to understand how it works.

As for the work directories, WORK_FOLDER is used to store the temporary VM the vboxmanage executable creates and manipulates before exporting the .ova file. It also seems to store some temporary files under /tmp that can cause issues (quark build at least consistently failed for me until I cleared the temp files)

meta-ostro/recipes-core/vboxmanage/vboxmanage-native_5.0.16.bb
```diff
@@ -0,0 +1,73 @@
...
+    --disable-extpack \
+    --disable-docs \
+    --disable-libvpx \
+    --disable-udptunnel \
+    --disable-devmapper \
+    --build-headless \
+    "
+
+do_configure() {
+    ./configure ${EXTRA_OECONF}
```
mythi: ```do_configure()``` should automatically run ./configure with the ```EXTRA_OECONF``` parameters so AFAIK ```do_configure()``` can be removed completely.

Valtis: I tried that; compilation failed.

pohly: On Wed, 2016-03-30 at 02:02 -0700, Erkka Kääriä wrote:
> > +do_configure() {
> > +    ./configure ${EXTRA_OECONF}
> 
> I tried that; compilation failed.

It's worth mentioning in the code why something unusual is needed, to
ensure that others know that it's intentional and also to document the
reason.

Valtis: After reading the doc (http://www.yoctoproject.org/docs/2.1/mega-manual/mega-manual.html#new-recipe-configuring-the-recipe), only autotools or cmake based projects handle do_configure() automatically. As this is neither, we must handle configuration manually. 

I added a comment to clarify this.

Valtis:
Yes, it's for retriggering testing. I'll keep the message in mind for the future.

Valtis:
retest this please

Valtis:
* Changed output name from foo.ova to foo.dsk.ova. The image.bbclass symlink creation now recognizes the image and creates a symlink for it automatically. 
* Due to above, removed manual symlink creation
* VM is now unregistered after exporting it. It seems that the host system Virtualbox installation would occasionally pick up the registered virtual machine and list it (not sure why this seems to be only occasional occurence). Unregistering removes it from the list, which avoids spamming the host Virtualbox VM list.

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -24,16 +24,93 @@
...
+  fi
+
+  # 64 bit linux with sata bus
+  OS_TYPE="Linux_64"
+  STORAGE_NAME="SATA"
+  STORAGE_BUS_TYPE="sata"
+
+
+  VM_NAME="${IMAGE_NAME}"
+  # 512 should be large enough value to run Ostro (Galileo has 256mb ram for example)
```
gvancuts: Nitpick... could you capitalize 'mb' and 'ram' (i.e. 256MB RAM)

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -24,16 +24,93 @@
...
+  # but small enough value that even a modest host machine can afford it
+  RAM="512"
+  # min allowed - as we do not run graphical environtment, it should be enough
+  VRAM="12"
+  # As above with ram, 2 cores should be enough for ostro without crippling the host machine
+  CPU_CORE_COUNT=2
+  # required for succesful boot
```
gvancuts: s/succesful/successful

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -24,16 +24,93 @@
...
+  STORAGE_BUS_TYPE="sata"
+
+
+  VM_NAME="${IMAGE_NAME}"
+  # 512 should be large enough value to run Ostro (Galileo has 256mb ram for example)
+  # but small enough value that even a modest host machine can afford it
+  RAM="512"
+  # min allowed - as we do not run graphical environtment, it should be enough
+  VRAM="12"
+  # As above with ram, 2 cores should be enough for ostro without crippling the host machine
```
gvancuts: s/ram/RAM

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -24,16 +24,93 @@
...
+  OS_TYPE="Linux_64"
+  STORAGE_NAME="SATA"
+  STORAGE_BUS_TYPE="sata"
+
+
+  VM_NAME="${IMAGE_NAME}"
+  # 512 should be large enough value to run Ostro (Galileo has 256mb ram for example)
+  # but small enough value that even a modest host machine can afford it
+  RAM="512"
+  # min allowed - as we do not run graphical environtment, it should be enough
```
gvancuts: s/environtment/environment

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -24,16 +24,93 @@
...
+  STORAGE_PORT="0"
+  STORAGE_DEVICE="0"
+  STORAGE_DEVICE_TYPE="hdd"
+
+  VIRTUALBOX_IMAGE="${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.dsk.vdi"
+  VIRTUALBOX_EXECUTABLE="${STAGING_BINDIR_NATIVE}/VBoxManage"
+
+  APPLIANCE_NAME="${VM_NAME}.dsk.ova"
+
+  FOLDER_NAME="virtualbox_vm"
```
gvancuts: Wouldn't it be more consistent with Yocto practices to call this folder virtualbox-vm?

Valtis:
* Typo fixes
* Changed the /tmp directory deletion to use find

As for the /tmp directory usage: It seems to be hardcoded (https://www.virtualbox.org/browser/vbox/trunk/src/libs/xpcom18a4/ipc/ipcd/shared/src/ipcConfig.cpp#L61). I could patch this to use some directory inside the Yocto build directory, but then we could face similar situation than we had with Java, where binaries grabbed from sstate fail on other build workers, as the build directory paths are different. This isn't also the only usage of /tmp, as 'grep -rIi "/tmp" | wc -l' in the source folder returns  429.

Valtis:
retest this please

Valtis:
retest this please

Minnowboard tests seem to have failed for unrelated reasons.

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -24,16 +24,91 @@
...
+  APPLIANCE_NAME="${VM_NAME}.dsk.ova"
+
+  FOLDER_NAME="virtualbox-vm"
+  WORK_FOLDER="${DEPLOY_DIR_IMAGE}/${FOLDER_NAME}"
+
+  mkdir -p ${WORK_FOLDER}
+
+  # erase any temp files from previous builds, as these can cause issues
+  # notably NS_ERROR_FACTORY_NOT_REGISTERED (0x80040154)
+  sudo find /tmp -name ".vbox-root*" -type d -prune -exec rm -rf {} \;
```
mythi: sudo isn't right. this can be modified such that you only search for files in /tmp/.vbox-error (I'm assuming you wanted to get of the 'permission denied' when scanning /tmp.)

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -24,16 +24,91 @@
...
+  ${VIRTUALBOX_EXECUTABLE} storagectl ${VM_NAME} --name ${STORAGE_NAME} --add ${STORAGE_BUS_TYPE}
+
+  # attach the .vdi image as hard drive
+  ${VIRTUALBOX_EXECUTABLE} storageattach ${VM_NAME} --storagectl ${STORAGE_NAME} --medium ${VIRTUALBOX_IMAGE} --port ${STORAGE_PORT} --device ${STORAGE_DEVICE} --type ${STORAGE_DEVICE_TYPE}
+
+  # ensure we are in correct working directory
+  cd ${DEPLOY_DIR_IMAGE}
+
+  # export the image
+  ${VIRTUALBOX_EXECUTABLE} export ${VM_NAME} --output ${APPLIANCE_NAME} --ovf20
```
mythi: can you avoid ```cd``` with ```--output ${DEPLOY_DIR_IMAGE}/${APPLIANCE_NAME}```

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -24,16 +24,91 @@
 #   This is required to identify it and pass its Partition UUID to the kernel, for booting.
 
 
-COMPRESSIONTYPES_append = " vdi"
+COMPRESSIONTYPES_append = " vdi ova"
```
mythi: there were some changes to this. please rebase and work with the conflicts

Valtis:
* Rebased on top of the latest meta-ostro
* Replaced the find exec rm - command with plain rm -rf, as the directory name is hardcoded and we can rely on it.
* --output now uses absolute path

mythi:
@Valtis Thanks! Looks +1 to me.

@igor-stoppa @jlaako can you please take a look

igor-stoppa:
If now we can have .ova files, why still produce the .vdi?
Btw, couldn't the ova use the .dsk directly?

I wouldn't hold back the patch only for what I mentioned, but I think it should be addressed in a couple of rework patches.

Valtis:
@igor-stoppa Virtualbox does not accept .dsk files as virtual hard drive (not recognized), so the .vdi is still required. If the file is not needed otherwise, then maybe the .vdi creation could be merged with .ova creation though.

igor-stoppa:
@Valtis are you sure? This seems to say otherwise: https://www.virtualbox.org/manual/ch09.html#rawdisk

The .vdi disk image creation should be left as option, I think.
But the creation of the .ova file should not require explicitly selecting the .vdi file.

Valtis:
@igor-stoppa VBoxManage: error: Could not get the storage format of the medium '/<snip>/ostro-image-intel-corei7-64-2016-04-07_23-35-47-build-315.dsk' (VERR_NOT_SUPPORTED)

kad:
Virtualbox should be able to work with "raw" images. and our dsk image is just raw image.
Virtualbox might be expecting raw images with some other extension, e.g. "*.img" or something similar.

Valtis:
Renaming the image to .img has no effect either. Using the page @igor-stoppa linked allows me to create .vmdk file from the .dsk file, and it boots succesfully. However the absolute path to the .dsk file on my machine is present in the .vdsk file, which makes me doubt it would work on other machines.

igor-stoppa:
@Valtis did you try to export the machine as appliance? I would expect this phase takes care of whatever might be installation specific, because appliances are meant to be importable wherever.

Valtis:
@igor-stoppa No, no I did not for some reason. Exporting it seems to work fine.

Valtis:
The appliance now builds its virtual hard drive from the raw image, so the vdi is no longer a dependency.

gvancuts:
IIRC from previous investigations, VirtualBox does not support raw images, they have to be converted to a supported format first (and VBoxManage supports converting from a raw image). Using VDI by default is what makes most sense given it's VirtualBox native format. It's still possible to convert that to anything else later (from the vdi or dsk image). I think it would make sense to stop publishing the vdi image if the ova appliance is functional. I would still keep the option in the code to be able to produce a vdi image too if desired.

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -38,6 +39,74 @@ inherit pythonnative
...
+  IOAPIC="on"
+
+  # default settings for hard drive
+  STORAGE_PORT="0"
+  STORAGE_DEVICE="0"
+  STORAGE_DEVICE_TYPE="hdd"
+
+  
+  RAW_IMAGE="${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.dsk"
+  VIRTUALBOX_IMAGE="${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.vmdk"
```
gvancuts: Why use VMDK (VMware native format) instead of VDI (VirtualBox native format) here?

Valtis: I just used the approach described here: https://www.virtualbox.org/manual/ch09.html#rawdisk

Looking at the available VBoxManage commands, converthd with -dstformat VDI might be more appropriate though.

However, if the vdi compression (conversion?) already creates a vdi file, and appliance creation creates a vdi file from the raw image, wouldn't this cause issues? At very least this step would be duplicated.

gvancuts: I'm not following the thinking to be honest. That VirtualBox manual page is really about how to give raw access to the host machine HDD. They seem to use an artefact from VMDK to achieve that.

You always need a disk image to be provided as part of an .ova appliance, creating a vmdk using an obscure feature of VirtualBox seems awkward given we have the ability to properly converting our .dsk into .vdi (and also given that the new steps don't really make us gain anything, we're still converting our .dsk image to something else).

What would make sense in my opinion is that we no longer *publish* .vdi images once .ova are available.

kad: created .vmdk file contains about 500 bytes of meta information about existing dsk file. so, on disk, space consumed only once. In case of conversion to vdi first, it means additional space used.

gvancuts: Doesn't that force us to publish an uncompressed .dsk as part of .ova? If so, that would be much larger than publishing an .ova that includes a .vdi file, wouldn't it?

kad: it might be if zip doesn't handle sparse files correctly. @Valtis  what is current sizes of different artifact ?

Valtis: the produced ova file is ~208M, which is the same it was when I was using the vdi file directly. This is considerably smaller than the .dsk image file, which is ~4 gigabytes, but if the image is mostly zeros, I guess the compression works well in this case.

kad: ova internally is zip file as I recall. How much space it actually uses once user downloads and adds ova to his VirtualBox ?

Valtis: About 584M.

Valtis:
retest this please

mythi:
This is very good now, IMO.

I'll merge it once if the builds/tests pass.

@Valtis we'd need another PR to update docs/local.conf.sample and drop *.vdi.* images

Valtis:
retest this please
Strange Jenkins error message.

I enabled the .ova generation in CI, which uncovered some issues with the VBoxManage recipe - turns out this pull request has never actually been tested properly in CI. There were some dependencies to libraries I had present on my machine. These libraries were not listed in the recipe DEPENDS variable or even whitelisted in some cases, and as such the build failed in CI.

Valtis:
Pull request was updated to (hopefully) fix the dependency issues. I can now compile it successfully on Ubuntu 14.04 in VM. Majority of the changes are just patching the config\build files to play nice with Yocto build system, but I had to add a new recipe for makeself-script, as it is a dependency and not provided by any of the existing layers as far as I can see.

Valtis:
I updated the pr to also remove the vdi-file generation and sha256sum generation for ova. The former is unnecessary and the latter causes the build to fail.

Valtis:
retest this please

No idea why I'm getting errors regarding nodejs runtime.

mythi:
@valtis there was a change in meta-ostro that needs an update in meta-iot-web too. That's why you see the failure

Valtis:
retest this please

meta-ostro/classes/image-dsk.bbclass
```diff
@@ -38,6 +39,74 @@ inherit pythonnative
...
+  #set core count, memory, firmware and ioapic
+  ${VIRTUALBOX_EXECUTABLE} modifyvm ${VM_NAME} --memory ${RAM} --vram ${VRAM} --cpus ${CPU_CORE_COUNT} --firmware ${FIRMWARE} --ioapic ${IOAPIC}
+
+  # add bus for hard drives
+  ${VIRTUALBOX_EXECUTABLE} storagectl ${VM_NAME} --name ${STORAGE_NAME} --add ${STORAGE_BUS_TYPE}
+  
+  # create virtual hard drive from the raw .dsk image
+  ${VIRTUALBOX_EXECUTABLE} internalcommands createrawvmdk -filename ${VIRTUALBOX_IMAGE} -rawdisk ${RAW_IMAGE}
+  
+  # attach the .vdi image as hard drive
```
gvancuts: We're now creating a vmdk image instead of vdi, aren't we? If so, can you update the comment to reflect that.

gvancuts:
Why does sha256sum generation for ova cause the build to fail? That sounds like something we should fix instead of removing altogether?

kad:
@gvancuts seems to be bug in image.bbclass in OE core that executes "compression" commands multiple times for the cases where it's not needed.

gvancuts:
Thanks @kad , is that bug being tracked/resolved elsewhere (as part of OE-core)? If so, then I'm OK dropping this for now.

kad:
Thanks to @Valtis bug now reported https://bugzilla.yoctoproject.org/show_bug.cgi?id=9482
Even proposed patch attached.

Valtis:
I fixed the comment issue. The script now also grants read permission for the output files, as they previously had permissions granted only for the owner. This caused a 403 error when attempting to download the files from the server.

Valtis:
retest this please

Soletta failure.

Valtis:
retest this please

Soletta related failure again.

mythi:
@Valtis please rebase and then I'll merge.

Valtis:
Tests failed due to a Galileo kernel panic. As this PR does not touch content on the image, I'd say it's unrelated. The previous, pre-rebase tests also passed just fine.

Nevertheless, retest this please

igor-stoppa:
@Valtis If you have still the log from the panic, can you please report it?

Valtis:
Sure. I'm also re-testing the image locally to see if I can reproduce it.

Valtis:
retest this please

Soletta failure again.

okartau:
retest this please

Valtis:
I renamed the EXTRA_OECONF variable to VBOX_CONF in the recipe, as the old name was misleading (we are not using the automake class), and more importantly, because the recent no-static-libs change added  --disable-static option into EXTRA_OECONF, which broke the configure step as Virtualbox configure script does not recognize this argument.

Valtis:
I just have the worst of luck when it comes to tests. This time Beaglebone flashing failed, as the file system was mounted as read-only mid-flashing. It may indicate that the SD card is failing. 

retest this please

Valtis:
I rebased the PR to see that all tests are still passing (previous result was good).

Valtis:
retest this please

Galileo failed to flash the image.

Valtis:
retest this please

Edison failed this time. Seems to be unrelated.

Valtis:
retest this please

Build machines failed to rsync the build results.

igor-stoppa:
@Valtis:
​This begs the question: should we hire an exorcist ​ ​for your PR or do also everything else fails so often so randomly?​

Valtis:
I think this PR would be a good candidate for most random failures before merge.

gvancuts:
+1 on hiring an exorcist ;-)

pohly:
Builds are finally good. I'll merge.

Regarding exorcists, I believe @Valtis is currently the most suitable person for the job. Kind of ironic that his PR was indeed one of those most affected by this ghost bugs in the testing ;-}

PR #84, opened 2016-03-24
local.conf.sample: add section to explain OSTRO_VM_IMAGE_TYPES

gvancuts:
Add a section that explains how to use the OSTRO_VM_IMAGE_TYPES to
tune the image formats being generated by bitbake. This commit adds
a little explanation but does not change the default behaviour. The
user will need to uncomment the line and set the variable to his
preference.

[skip ci]


kad:
:+1:

mythi:
@gvancuts with my comment about "but not limited to" I was referring to "OSTRO_VM_IMAGE_TYPES = "dsk.xz dsk.zip dsk.vdi.zip" set in ostroproject-ci.inc. These could be added to local.conf.sample too.

We don't need to mention the other compressiontypes you found by grepping the source as we've not tried/tested them.

gvancuts:
Rebased to latest meta-ostro/master and force-pushed an update that addresses all comments.

PR #85, opened 2016-03-24
ostro.conf: disable running fsck at boot time

jlaako:
Disable running fsck at boot. System clock is typically wrong at early
boot stage due to lack of RTC backup battery. This causes unnecessary
fixes being made due to filesystem metadata time stamps being in future.

Ostro images use journaling filesystem and thus running fsck at boot is
not necessary


pohly:
Will a similar change be added for the Edison BSP or is it not necessary there? If I understand it correctly, the Edison BSP does not use APPEND.

Other than that the change makes sense, so +1.

ipuustin:
Looks good to me, +1. This might make boot faster too.

However, if we want to, we can also make fsck to not consider the timestamp difference. This requires either an environment variable to fsck (passed via systemd-fsck) or having a configuration file in `/etc`.

mythi:
Also aligned with what was discussed earlier, +1

PR #87, opened 2016-03-31
local.conf.sample: enable SSTATE_MIRRORS by default

kad:
Ostro Project CI produces uninative SSTATE cache which is
beneficial for most of users to speed-up builds.
Enable it by default in local.conf

### DIFF
--- a/doc/howtos/building-images.rst
+++ b/doc/howtos/building-images.rst
@@ -174,8 +174,8 @@ parts do not need to be rebuilt. The Yocto Project shared state code supports
 incremental builds and attempts to accelerate build time through the use
 of prebuilt data cache objects configured with the ``SSTATE_MIRRORS`` setting.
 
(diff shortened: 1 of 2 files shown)


barbieri:
+1

FINALLY :-) do we need to update the rst that documents it should be enabled?

kad:
@dbkinder as I recall already described it. maybe we need to update few words about having it enabled by default and how to disable.

dbkinder:
Just did in your git repo ☺
-- david

kad:
Gerat! Thanks @dbkinder, merged, pushed updated version.

mythi:
@kad this changesjustifies CONF_VERSION and LOCALCONF_VERSION bump

kad:
not sure. not really critical in my opinion, we anyway would have bumps very soon due to swupd

mythi:
test this please

(basically no-op but doing it anyway)

PR #89, opened 2016-04-05
devtool: add kernel-menuconfig subcommand

bluelightning:
Add a kernel-menuconfig subcommand (only accessible within the
extensible SDK) that first puts the kernel into the workspace if it
isn't already, and then launches "make menuconfig".

(This is being added on the Ostro side because we'll need to do some
significant refactoring before it could go into OE-Core).

Signed-off-by: Paul Eggleton <paul.eggleton@linux.intel.com>

### DIFF
--- a/meta-ostro/lib/devtool/__init__.py
+++ b/meta-ostro/lib/devtool/__init__.py
@@ -1,0 +1,3 @@
+# Enable other layers to have modules in the same named directory
+from pkgutil import extend_path
+__path__ = extend_path(__path__, __name__)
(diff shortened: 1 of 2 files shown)


bluelightning:
I should mention this requires http://cgit.openembedded.org/openembedded-core/commit/?id=626dbadf22b57a22a8f8b9d1957937120f4ba4d5

mythi:
@bluelightning thanks

we're still missing the patch but it's queued for integration. I'll trigger a test build once that is included.

mythi:
accepting to CI tests since the oe-core patch is now merged.

ok to test

mythi:
restarting the test...

test this please

mythi:
@bluelightning could you rebase this commit please

PR #93, opened 2016-04-06
ostroproject-ci: enabler for flexible automatic testing

kad:
New variable OSTROPROJECT_CI_TEST_RUNS allows
to configure which images and with which test suite
and test data would be tested after CI build completed.
Format is space separated list of tuples.
Each tuple is comma separated list of items:
- image name to test
- test suite archive for that image
- test data archive for testsuite
- MACHINE build target

Variable can be overriden for each individual MACHINE


mythi:
This makes sense and is aligned with what is being discussed on the mailing list. The only open is how to get meta-iotqa to provide the right test suites for each image.

PR #94, opened 2016-04-07
ostro-image.bb: removed unused app-privileges and added app-framework package group.

klihub:
Added a package group for app framework. The app framework package group is now unconditionally included in the image as requested by QA. Removed the old and now unused app-privileges package group that used to collectively pull in the cynara, security manager, and app framework. Also removed the image post-processing workaround for the security-manager sqlite DB.

### DIFF
--- a/meta-ostro/classes/ostro-image.bbclass
+++ b/meta-ostro/classes/ostro-image.bbclass
@@ -56,11 +56,11 @@ OSTRO_IMAGE_PKG_FEATURES = " \
 #
 # swupd = install swupd client and enabled generation of swupd bundles
 IMAGE_FEATURES[validitems] += " \
-    app-privileges \
     autologin \
     ima \
     smack \
     swupd \
+    app-framework \
     ${OSTRO_IMAGE_PKG_FEATURES} \
 "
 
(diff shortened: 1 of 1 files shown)


pohly:
Is it really still valid that QA needs AppFW installed by default? What's the argument for needing that?

The rest of the changes look good, but I'd prefer to not merge this yet, because it conflicts with the on-going meta-swupd integration.

klihub:
QAs argument for having app-framework in by default was that they want to be able to run test cases with an unmodified image. They would like to be able to put their test cases (test applications) on top of the image without having to build one for testing. I doubt this is an entirely valid point as I suspect they will need to have testing-specific additions in the image anyway. If you'd prefer to merge later, once the swupd conflict situation is gone, without the unconditional inclusion bits, just let me know and I modify the PR accordingly.

mythi:
@klihub iot-app-fw is currently pulled in via example-* apps QA directly adds to the image.

mythi:
As requested, let's take it after meta-swupd work. Adding labels.

klihub:
@mythi Yup. So I guess your author-rework-required label is trying to say I should remove the unconditional inclusion bits, right ?

klihub:
Updated to a single squashed patch which adds the new app-framework feature (not enabled by default), and removes the old app-privileges feature.

mythi:
On Thu, Apr 7, 2016 at 3:25 PM, Krisztian Litkey <notifications@github.com>
wrote:

> @mythi <https://github.com/mythi> Yup. So I guess your
> author-rework-required label is trying to say I should remove the
> unconditional inclusion bits, right ?
>
>
@klihub rework would also be needed since this is taken after swupd stuff.
That changes this file quite a bit.

klihub:
@mythi Rebased and reworked on top of the latest branch with the swupd bits included.

mythi:
@klihub thanks!

I'll trigger a new test build. 

test this please

PR #95, opened 2016-04-07
meta-swupd

pohly:
This completes the integration of meta-swupd. The PR is ready for merging.


kad:
:-1: 
```
do_swupd_update_prepend ()
```
must be removed.

pohly:
On Thu, 2016-04-07 at 05:58 -0700, Alexander D. Kanevskiy wrote:
> do_swupd_update_prepend ()
> 
> must be removed.

Can't test that here, so I triggered a test build with such a change
included in ostro-os PR #25.

kad:
We anyway would need to re-test this PR after ostroproject/ostro-os#51 merged.
There is no harm of updates inside meta-swupd at the moment, it wouldn't block merge there.
But this PR, which enables usage of meta-swupd for real, this must be tested before merging.

pohly:
Yes, this PR needs to be tested. @mythi please merge https://github.com/ostroproject/ostro-os/pull/51 as soon as possible.

If the test build in PR #25 passes, then I will add that additional commit also here.

meta-ostro/classes/ostro-image.bbclass
```diff
@@ -10,121 +10,245 @@ IMAGE_INSTALL = " \
...
     iotivity \
     java-jdk \
     node-runtime \
     nodejs-runtime-tools \
     python-runtime \
     qatests \
-    smack \
+    ssh-server \
     soletta \
     soletta-tools \
```
barbieri: what is soletta-tools x soletta? In this image only installing the sol-fbp-runner is enough, it should bring the dependencies such as other libraries and modules. and I'd call it similar to other runtimes (python-runtime, java-runtime, soletta-runtime).

For -dev images, installing soletta-dev, which should include sol-fbp-generator and other scripts would be necessary.

Last but not least, java-jdk, node-runtime and nodejs-runtime-tools, can we have a single name pattern?

meta-ostro/classes/ostro-image.bbclass
```diff
@@ -10,121 +10,245 @@ IMAGE_INSTALL = " \
...
+# in building-images.rst.
+#
+# swupd = install swupd client and enabled generation of swupd bundles
 IMAGE_FEATURES[validitems] += " \
-    ptest-pkgs \
-    ssh-server-openssh \
-    tools-debug \
-    tools-profile \
+    app-privileges \
+    autologin \
```
barbieri: ok for now, but in future this should be moved to the "application" that will provide login, so we can avoid the login stuff in our host, only inside the container.

meta-ostro/classes/ostro-image.bbclass
```diff
@@ -10,121 +10,245 @@ IMAGE_INSTALL = " \
...
+"
+
+# When swupd bundles are enabled, choose explicitly which images
+# are created. The base image will only have the core-os bundle and
+# thus might not be very useful. For use in ostro-image-swupd.bb we
+# pre-define additional images:
+# ostro-image-swupd-reference -
+#    Base image plus login via getty and ssh, plus connectivity.
+#    This is what developers  are expected to start with when
+#    building their first image.
```
barbieri: This is good and should be in our os-core, no need to be extra bundles of our reference image.

Products would define their own bundles and those would be free to define more fine grained options in os-core, like just bluez but not connman. That is: base image for Ostro-Reference == os-core only. This is the "command line reference image".

meta-ostro/classes/ostro-image.bbclass
```diff
@@ -10,121 +10,245 @@ IMAGE_INSTALL = " \
...
+# it does not parse the output correctly.
+#
+# For now avoid these problems by sticking to the traditional
+# mount utilities from util-linux.
+export ALTERNATIVE_PRIORITY_UTIL_LINUX ?= "305"
+
+# We do not know exactly which util-linux packages will get
+# pulled into bundles, so we have to install all of them
+# also in the os-core. Alternatively we could try to select
+# just mount/umount as overrides for Toybox/Busybox.
```
barbieri: nice catch, but if we can patch toybox to provide a more complete and compliant `mount`,  this can be avoided. Ask the team that does toybox, maybe they can solve this in parallel.

meta-ostro/classes/ostro-image.bbclass
```diff
@@ -156,7 +280,21 @@ FEATURE_PACKAGES_soletta-tools = "soletta-dev-app"
 # because it is the most common source code management tool.
 FEATURE_PACKAGES_tools-develop = "packagegroup-core-buildessential git"
 
-FEATURE_PACKAGES_swupd = "packagegroup-swupd"
+# Add bash because it is more convenient to use than dash.
+# This does not change /bin/sh due to re-organized update-alternative
+# priorities.
+FEATURE_PACKAGES_tools-interactive = "packagegroup-tools-interactive bash"
```
barbieri: please add more useful helpers for interactive use, like bash-completion? :-)

I'm thinking the case where we have all daemons split in command line clients and actual server, in that case should we add the cmdline clients here? like connmanctl, systemctl, systemd console agents... It would be nice to make this more automatic, so cmdline tools split packages can be auto pulled in.

nothing to be done now, but let's think about it.

meta-ostro/classes/ostro-image.bbclass
```diff
@@ -156,7 +280,21 @@ FEATURE_PACKAGES_soletta-tools = "soletta-dev-app"
...
+
+# We could make bash the login shell for interactive accounts as shown
+# below, but that would have to be done also in the os-core and thus
+# tools-interactive would have to be set in all swupd images.
+# TODO (?): introduce a bash-login-shell image feature?
+# ROOTFS_POSTPROCESS_COMMAND_append = "${@bb.utils.contains('IMAGE_FEATURES', 'tools-interactive', ' root_bash_shell; ', '', d)}"
+# root_bash_shell () {
+#     sed -i -e 's;/bin/sh;/bin/bash;' \
+#        ${IMAGE_ROOTFS}${sysconfdir}/passwd \
+#        ${IMAGE_ROOTFS}${sysconfdir}/default/useradd
```
barbieri: when we move interactive stuff into a container with its own stuff, we can define the root login inside the container to use /bin/bash. Actually bash can be provided only in that container.

meta-ostro/classes/ostro-image.bbclass
```diff
@@ -348,6 +485,27 @@ ima_evm_sign_rootfs_prepend () {
 # for that machine.
 APPEND_append = "${@bb.utils.contains('IMAGE_FEATURES', 'smack', '', ' security=none', d)}"
 
+# In addition, when Smack is disabled in the image but enabled in the
+# distro, we strip all Smack xattrs from the rootfs. Otherwise we still
+# end up with Smack labels in the filesystem although we neither need
+# nor want them, because the packages that were compiled for the distro
+# have Smack enabled and will set the xattrs while getting installed.
```
barbieri: do they cause any problems? why not leave them even if unused?

meta-ostro/classes/ostro-image.bbclass
```diff
@@ -404,3 +562,18 @@ python extra_motd () {
...
 ROOTFS_POSTPROCESS_COMMAND += "${@'extra_motd;' if d.getVar('OSTRO_EXTRA_MOTD', True) else ''}"
+
+# Ensure that the os-release file contains values matching the current image creation build.
+# We do not want to rebuild the the os-release package for that, because that would
+# also trigger image rebuilds when nothing else changed.
+ostro_image_patch_os_release () {
+    sed -i \
+        -e 's/distro-version-to-be-added-during-image-creation/${DISTRO_VERSION}/' \
+        -e 's/build-id-to-be-added-during-image-creation/${BUILD_ID}/' \
+        ${IMAGE_ROOTFS}/usr/lib/os-release
```
barbieri: maybe also the `NAME=`, `ID=`, `PRETTY_NAME=`, `HOME_URL=`, `SUPPORT_URL=` and `BUG_REPORT_URL=`?

meta-ostro/conf/conf-notes.txt
```diff
@@ -1,10 +1,7 @@
 Common targets are:
-    ostro-image
-    ostro-image-dev
-    core-image-minimal
-    meta-toolchain
-    adt-installer
-    meta-ide-support
+    ostro-image-noswupd (when building without swupd,
+                         the recommended mode for local image building)
+    ostro-image-swupd-reference (when building with swupd)
```
barbieri: "the recommended mode for deployment"?

meta-ostro/conf/distro/include/ostro-os-development.inc
```diff
@@ -1,8 +1,10 @@
 # Use the pre-generated keys for IMA signing.
 IMA_EVM_KEY_DIR = "${IMA_EVM_BASE}/data/debug-keys"
 
-# Enable local root access.
-OSTRO_EXTRA_IMAGE_FEATURES += "autologin"
+# Enable local root access. Always use _append, to
+# avoid getting this change overwritten by a
+# OSTRO_IMAGE_EXTRA_FEATURES = "foo" in local.conf.
+OSTRO_IMAGE_EXTRA_FEATURES_append = " autologin"
```
barbieri: as before, we can keep going like this to get it merged. But future should have images where login are confined to a container.

meta-ostro/conf/distro/ostro.conf
```diff
@@ -12,6 +12,52 @@ SDK_UPDATE_URL = "\
...
 
+# When building in a CI system with swupd enabled, OS_VERSION must be
+# set to a consistent value for all builds. See ostroproject-ci.inc
+# for an example how that works with Jenkins.
+#
+# In local builds, we only have ${DATETIME} as something that
+# increments automatically, but it is too large for an integer number
+# on 32 bit systems. Therefore we substract the 2016 as the initial
+# year in which Ostro OS started using swupd (assumes a 64 build host,
+# to avoid integer overflows in Python itself) and ignore the
```
barbieri: there are no integer overflows in python, it will auto-grow to large number, which can contain more than simple 32/64 bits :-)

meta-ostro/conf/distro/ostro.conf
```diff
@@ -12,6 +12,52 @@ SDK_UPDATE_URL = "\
 		${MACHINE}\
 "
 
+# When building in a CI system with swupd enabled, OS_VERSION must be
+# set to a consistent value for all builds. See ostroproject-ci.inc
+# for an example how that works with Jenkins.
+#
+# In local builds, we only have ${DATETIME} as something that
+# increments automatically, but it is too large for an integer number
+# on 32 bit systems. Therefore we substract the 2016 as the initial
```
barbieri: maybe ignore second precision and go with `%Y%m%d%H%M`?

meta-ostro/recipes-core/packagegroups/packagegroup-tools-interactive.bb
```diff
@@ -0,0 +1,30 @@
...
+    iputils-ping \
+    iputils-ping6 \
+    iputils-tracepath \
+    iputils-tracepath6 \
+    iputils-traceroute6 \
+    procps \
+    rsync \
+    traceroute \
+    vim \
+    wget \
```
barbieri: replace with curl? or add curl?

meta-ostro/recipes-swupd/clr-systemd-config/files/0002-systemd-service-for-updating-the-efi-combo.patch
```diff
@@ -1,40 +0,0 @@
...
---- /dev/null
-+++ b/system/efi-combo-trigger.service
-@@ -0,0 +1,7 @@
-+[Unit]
-+Description=Updates the EFI combo application
-+BindsTo=update-triggers.target
-+
-+[Service]
-+Type=oneshot
-+ExecStart=/usr/bin/efi_combo_updater
```
barbieri: should this be in some hidden namespace @igor-stoppa? we do not need this to end in $PATH, do we?

barbieri: sorry, this file was removed, I just looked at the green part of the github highlight :-)

meta-ostro/recipes-swupd/swupd-client/swupd-client/efi-combo-trigger.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Updates the EFI combo application
+BindsTo=update-triggers.target
+
+[Install]
+WantedBy=update-triggers.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/efi_combo_updater
```
barbieri: should this tool be in hidden namespace so it doesn't end in $PATH? @igor-stoppa 

also missing newline :-)

meta-ostro/recipes-swupd/swupd-client/swupd-client_%.bbappend
```diff
@@ -0,0 +1,38 @@
...
+
+# Optionally add our efi-combo-trigger.
+SYSTEMD_SERVICE_${PN} += "${@ 'efi-combo-trigger.service' if ${OSTRO_USE_DSK_IMAGES} else ''}"
+
+# And activate it.
+SYSTEMD_AUTO_ENABLE_${PN} = "enable"
+
+do_compile_append() {
+    if [ "${OSTRO_USE_DSK_IMAGES}" = "True" ]; then
+        ${CC} ${LDFLAGS} ${WORKDIR}/efi_combo_updater.c  -Os -o ${B}/efi_combo_updater `pkg-config --cflags --libs glib-2.0`
```
barbieri: I don't see any glib being used :-)

scripts/swupd-http-server
```diff
@@ -0,0 +1,64 @@
...
+fi
+if [ $count -gt 1 ]; then
+    echo >&2 "$dir is ambiguous. Specify one as parameter:"
+    ls >&2 -d -1 $dir
+    exit 1
+fi
+
+echo "Serving swupd files from" $dir
+cd $dir
+exec -a swupd-http-server python -m SimpleHTTPServer $port
```
barbieri: god, why not do it all in python? It's the less effort than sh + py :-)

barbieri:
hi @pohly,

Overall I like it and would recommend to include ASAP. I did some minor comments, some are not even related to your PR (like efi_combo_updater, which I've created an issue https://github.com/ostroproject/meta-ostro/issues/96).

Those comments about interactive and shell can be reviewed later, as we clean some parts the others will become cleaner.

meta-ostro/recipes-swupd/swupd-client/swupd-client/efi-combo-trigger.service
igor-stoppa: On 7 April 2016 at 10:01, Gustavo Sverzut Barbieri <notifications@github.com
> wrote:

> In
> meta-ostro/recipes-swupd/swupd-client/swupd-client/efi-combo-trigger.service
> <https://github.com/ostroproject/meta-ostro/pull/95#discussion_r58907078>
> :
>
​[...]​

> > +ExecStart=/usr/bin/efi_combo_updater
>
> should this tool be in hidden namespace so it doesn't end in $PATH?
> @igor-stoppa <https://github.com/igor-stoppa>
>
> also missing newline :-)
>
​yeah, however I'm not happy of the overall implementation
I have a temporary patch that turns this into a shell script, which is
cleaner
and doesn't rely on mount, however what I really want is to agree with
swupd guys
to introduce the concept of generic update-boot-files

I had some exchange with Arjan but the discussion kinda died out.
Your comments there would be appreciated.​



-- 
igor

barbieri: ok, as it is it's better to move to shell. we can review that (I filled an issue as per above) later.

on clear/arjan, I'd say he agreed with you, so send an ACK with actual changes, maybe a patch/PoC. He says basically "this is the case today", so if we miss a small detail, let them know.

pohly:
test this please

pohly:
The latest build in https://github.com/ostroproject/meta-ostro/pull/25 with symlink workaround removed failed on Gigabytes and Minnowmax because of ssh connectivity issues between test host and test devices. Not sure whether that was related to anything in the PR itself.

In this PR, I kicked of two builds: https://ostroproject.org/jenkins/view/Pull-Requests/job/dsl-meta-ostro_pull-requests/200/ with the symlink workaround, https://ostroproject.org/jenkins/view/Pull-Requests/job/dsl-meta-ostro_pull-requests/201/ with the workaround removed via commit 15aa34d.

meta-ostro/classes/ostro-image.bbclass
pohly: In this PR, I just followed the existing naming of image features. Let's clean this up separately.

meta-ostro/classes/ostro-image.bbclass
pohly: Let's see how that "shell application" goes. Core OS developers will always need the ability to investigate things in the core OS; if the "shell application" can provide that, then we can remove this direct login.

meta-ostro/classes/ostro-image.bbclass
pohly: So following that logic, ostro-image.bbclass as the base class for custom images shouldn't contain any bundle definitions, right? That would be in the OEM's swupd image, which would be ostro-image-swupd-reference.bb in our case.

If we change the base name like this, the main image will still be called "ostro-image-swupd-reference". But bundle images with additional bundles become ostro-image-swupd-reference-<foo>, for example ostro-image-swupd-reference-dev. Is that a good or bad thing?

I'm undecided myself. It's consistent and keeps the image name space clean, but it's also a bit long.

meta-ostro/classes/ostro-image.bbclass
pohly: We don't have all (any?) of those. But you are right, this can be enhanced to handle more fields, should they exist in os-release => separate PR, though.

meta-ostro/conf/conf-notes.txt
pohly: Fine with me. I just need to figure out where to put that change.

meta-ostro/conf/distro/ostro.conf
pohly: I had wondered about that and checked docs, but misread the explanation. Anyway, as it works automatically, then the code is fine and just the comment exposes my misunderstanding ;-}

If I get  the chance (= we need to retest for some other reason), I'll clean it up, otherwise do it later.

meta-ostro/conf/distro/ostro.conf
pohly: The current code already ignores seconds. But %Y%m%d%H%M is still to large for signed 32 bit, hence the substraction of the start year.

meta-ostro/recipes-core/packagegroups/packagegroup-tools-interactive.bb
pohly: Curl is already in the RDEPENDS. wget is present because meta-iotqa uses it in some scripts. I find it useful to provide both because users might prefer one or the other. wget also is a good test case that the merged CA certificate file works.

meta-ostro/recipes-swupd/swupd-client/swupd-client_%.bbappend
pohly: I expect that @igor-stoppa will cover that as part of updating the EFI combo updater.

scripts/swupd-http-server
pohly: The usual: it was a one-line wrapper adding the required parameters, and then it grew. I swear, it did that all by itself, I had nothing to do with it! ;-}

That's just an explanation, but no excuse, of course. Really should be pure Python.

meta-ostro/classes/ostro-image.bbclass
mythi: To answer to @barbieri: soletta-tools is simply soletta-dev-app. soletta in turn is what soletta.bb gives (the package group used to have more packages but they are now pulled in with RDEPENDS)

PR #97, opened 2016-04-08
soletta-dev-app: fix dangling symlinks in rootfs

mythi:
dev-app's do_install copies everything from ${S}, including
patches/ directory that is created by bitbake.

patches/ includes symlinks to patches applied from SRC_URI. These
symlinks point to destinations that do not get included in
do_install.

ostro-sanity.bbclass checks for dangling symlinks in rootfs as to
ensure swupd-enabled images work correctly.

To fix the QA errors, remove patches/ before copying everything
from ${S}.


mythi:
test this please

"Slave went offline"

PR #99, opened 2016-04-11
doc: added a vulnerability management document.

ipuustin:
The document gives information about how Ostro Project handles security vulnerabilities. Note that the document is still work in progress: the email addresses and mailing lists mentioned are not in use at this point of time. If the document passes review, the addresses, keys etc. should be activated before merging the PR.


ereshetova:
@cschaufler, @terriko  please take a look

doc/howtos/vulnerability-management.rst
```diff
@@ -0,0 +1,145 @@
...
+============
+
+Software vulnerabilities are found continuously even in mature pieces
+of software. This means that one of the key objectives of a Linux
+distribution is pushing fixes to vulnerabilities down to the end users.
+Ostro OS images will not be used as-is by end users, but Ostro OS
+will be the base software distribution for many products the Ostro OS
+customers are developing. These products must be kept secure for their
+expected lifetime. The published Ostro OS images are for development
+purposes only and should not be used in real products.
```
terriko: Should probably have a sentence or two here describing how real products should be deployed with Ostro so readers know how it differs from the Ostro images without having to search for this information themselves.  Something like "The development images do not auto-update.  Deployed products should be built from Ostro sources and contain a method for updating as needed; see $link for more details on the expected path for deploying Ostro in products."

doc/howtos/vulnerability-management.rst
```diff
@@ -0,0 +1,145 @@
...
+because not all CVE issues are exactly mapped to packages. For those
+components, such as Linux kernel, CVEs and vulnerability reports have
+to be investigated separately. Also, cve-check-tool doesn't report
+vulnerabilities that are not found in the vulnerability database.
+
+When Ostro OS -specific vulnerabilities are found that don't have a CVE
+entry, the person finding them can notify the Ostro OS Security Team by
+reporting the vulnerability using the secure contact email address and
+public key for communication encryption.
+
```
terriko: Link to the page with the email and public key info here.  I didn't see it at a glance from https://ostroproject.org/#get-involved so it may need to be created and linked appropriate there alongside the bug reporting link.

doc/howtos/vulnerability-management.rst
```diff
@@ -0,0 +1,145 @@
...
+reports. The correct procedure is to analyze the issue and then file a
+report to Ostro OS issue tracker if required.
+
+An important consideration is that Ostro Project does not catch the
+vulnerabilities in third-party software that our customers might be
+using. These customers need to track the potential security
+vulnerabilities in those software components themselves or rely on the
+Yocto project or other upstream meta-layer maintainers to handle the
+detection and fixing of the vulnerabilities.
+
```
terriko: I don't think "third-party" is clear enough here, since readers of this document may be consumers who consider Ostro OS to be third party code within their models.  Maybe "software not part of the core Ostro OS but deployed by customers" shortened to "non-core software" ?

doc/howtos/vulnerability-management.rst
```diff
@@ -0,0 +1,145 @@
...
+When a new vulnerability in Ostro OS or in a component used by Ostro OS
+is detected, the first step is to assess if the vulnerability affects
+Ostro OS. This is done by examining the vulnerability report, Ostro OS
+component version, and potentially running any proof-of-concept
+exploits or vulnerability checks distributed with the vulnerability
+report. It is possible that the Ostro OS security mechanisms, such as
+minimal configuration of system services, already prevent the
+vulnerability from directly affecting Ostro OS. If that’s the case, the
+criticality of the vulnerability will naturally be lower.
```
terriko: "criticality" should be changed to "risk level" to match up more clearly with the terminology used for security assessment.

doc/howtos/vulnerability-management.rst
```diff
@@ -0,0 +1,145 @@
...
+Finally, the processing of the vulnerability leads to a (periodic) look
+at Ostro OS security architecture. The Ostro Project security team has
+to consider if Ostro OS security architecture needs be adapted to
+better protect against the vulnerability and others like it. This is
+continuous work and eventually lead to an Ostro OS security
+architecture that is well prepared to handle many vulnerability
+classes.
+
+Tracking and discussing the security vulnerability and the fix to it
+happens in Ostro OS issue tracker. The security vulnerabilities in the
```
terriko: English weird on that first sentence.  Should probably be "Tracking and discussion of the security vulnerability and its fix will happen in the Ostro OS issue tracker."

dbkinder:
seems like this document is a companion to the other two security documents we have over in the architecture section (rather than a "how to" kind of document).  When we're ready to incorporate this, I'd recommend moving it over to the architecture folder.

doc/howtos/vulnerability-management.rst
ipuustin: As mentioned in the PR text, there isn't yet a key or email account in existence. If the document passes review, let's create those before merging.

ipuustin:
I did (most of) the changes suggested by @terriko and moved the document to `doc/architecture` directory.

mythi:
[skip ci]

mythi:
+1. I set the labels to indicate additional reviews/rework still needed.

doc/architecture/vulnerability-management.rst
```diff
@@ -0,0 +1,146 @@
...
+=================================
+
+The Ostro Project doesn't provide binaries to end users, meaning people
+who run devices which have Ostro OS -based distributions. Instead the
+Ostro OS customers are ODMs, which create and distribute binaries to the
+devices. For this reason, it’s paramount that the ODMs actively track
+Ostro OS security issues and also deliver security updates to their end
+users via the normal software update process.  Ostro OS is an
+open-source project, so Ostro Project itself does not necessarily know
+which ODMs actually use Ostro OS code in their devices.
```
ereshetova: Based on what we discussed today in SDL call, Ostro will provide updates to their published os images that users can pull manually. I think this must be also listed there and told that in case of severe vulnerabilities binary updates can be pulled into images using certain command line sequence.

ereshetova: So, you need to reflect two points in the doc: 1) images that are published are not for products and should not be used for products (this part is present now) and 2) however, ostro is nice enough to publish updates for these images if one wants to pull them manually for some reason. How often will updates be published? right when any component is changed in master or only in case we have significant vulnerability? I think this needs to be reflected also.

doc/architecture/vulnerability-management.rst
```diff
@@ -0,0 +1,146 @@
+Ostro OS vulnerability handling
```
dbkinder: add this as first line of the file (this will make inter-document linking to this document possible):

.. _vulnerability-management:

dbkinder:
After this get's through the technical review and merged, I'd like to go through it for some clarity editing that's probably easier to edit directly rather than suggesting changes.

doc/architecture/vulnerability-management.rst
ipuustin: I added a note about the software update for development images in a new document version.

ipuustin:
@mythi, @dbkinder: are we waiting for author rework or techwrite review anymore with this document? Can we remove the tags if not?

dbkinder:
This note mentions mailing list that doesn't exist and needs to before the document goes live.  Actually, if it's a mailing list, it would be ostro-security@lists.ostroproject.org instead of the an ostro-security@ostroproject.org as written.  

There's mention of "secure contact email address and
public key for communication encryption" that's not setup either, right?

Before merging we'll also need to add add a line to architecture.rst to reference this doc in the TOC

dbkinder:
Looks like only an admin can add or remove the author-rework-required and techwriter-review-required labels.  Other than mentioning mailing lists and email addresses that aren't set up yet, I think this doc is ready to go.  (I can't update the architecture.rst file to add it to the index because @mythi has a PR in progress on that doc.)

dbkinder:
is this needed in the 1.0 release?

ipuustin:
I've requested the mailing lists, addresses and keys (hi @kad!).

PR #100, opened 2016-04-12
doc: cover changes related to swupd image reorganization

pohly:
@dkinder @gvancuts please help review. This completes the integration
of meta-swupd as far as normal developers go. Documenting the usage of
swupd and swupd-based images is intentionally left out for now. I don't
have plans to add anything further about that.

This also replaces PR #98.

[skip ci]

PR #101, opened 2016-04-12
enhance development/production error message

pohly:
@uhofemei is this better? :-) Also note that PR #100 enhances building-images.rst such
that the need to choose the mode is mentioned more prominently.


kad:
:+1:

PR #104, opened 2016-04-12
simplify image content and bundles

pohly:
Removes the extra "ostro-image-swupd-reference" image and the
"reference" bundle and instead moves the content into the base images
of both ostro-image-swupd and ostro-image-noswupd.

Because the doc changes would conflict with the changes in PR #100,
this PR here includes all of those commits, too. That way we can
merge PR #100 and then this one, or this one directly.

@barbieri I hope this addresses your review feedback from PR #95.

@olev @mikko @kad I expect this to speed up the builds a bit.

### DIFF
--- a/doc/howtos/building-images.rst
+++ b/doc/howtos/building-images.rst
@@ -247,14 +247,12 @@ Alternatively, ``CORE_IMAGE_EXTRA_INSTALL`` can also be used. The
 difference is that this will also affect the initramfs images, which is
 often not intended.
 
-Including ``ostro-os-development.inc`` will automatically extend the
-configuration of ``ostro-image-noswupd`` such that the content matches
-what gets published as the ``ostro-image-swupd-reference``. For
-example, an ssh server gets added. If that is not desired, uncomment
-the following lines in ``conf/local.conf``::
+The example ``ostro-image-swupd`` is defined such that its default
+content corresponds to ``ostro-image-noswupd``. It is possible to
+reconfigure it so that it matches ``ostro-image-swupd-dev``::
 
-  OSTRO_DEVELOPMENT_EXTRA_FEATURES = ""
-  OSTRO_DEVELOPMENT_EXTRA_INSTALL = ""
+    OSTRO_IMAGE_NOSWUPD_EXTRA_FEATURES_append = "${OSTRO_IMAGE_FEATURES_DEV}"
+    OSTRO_IMAGE_NOSWUPD_EXTRA_INSTALL_append = "${OSTRO_IMAGE_INSTALL_DEV}"
 
 
 Accelerating Build Time Using Shared-State Files Cache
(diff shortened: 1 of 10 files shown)


doc/howtos/building-images.rst
```diff
@@ -78,6 +81,51 @@ the existing user accounts (including root) has a password set, so
...
+with ``# require conf/distro/include/ostro-os-development.inc`` and
+uncomment it. This will also add some recommended software to the ``ostro-image-noswupd``
+reference image, see below for details.
+
+
+Production Images
+-----------------
+
+When building production images, first follow the instructions
+provided in :file:`meta-intel-iot-security/meta-integrity/README.md` for creating your own
```
barbieri: @dbkinder and @gvancuts should we convert that doc to `.rst` and then integrate it in our website autogen documentation?

doc/howtos/building-images.rst
```diff
@@ -171,24 +194,66 @@ Example::
...
 
-To do this, before building,  edit the :file:`conf/local.conf` configuration file, 
-find the line
-with ``# require conf/distro/include/ostro-os-development.inc`` and
-uncomment it.
+Developers are encouraged to start building images the traditional way
+by using image recipes like ``ostro-image-noswupd`` where swupd is
+turned off, because:
+
+a) swupd support is still new and may have unexpected problems.
```
barbieri: I'd not write this in the documentation :-D

doc/howtos/building-images.rst
dbkinder: Yes, this should be included in the documentation set.

README files are OK for some explanatory information about the component, but we should not have users wandering around components looking for them.  We (Geoffroy/David) should take a look around the other components and see if there are substantial gems of information scattered around in README files we can incorporate into the documentation set.


-- david

doc/howtos/building-images.rst
pohly: But it's true, so why not call the child by its name? ;-} Anyone claiming that swupd support will be rock-solid right after integrating it is, well, very optimistic.

If anyone has suggestions for a better working of that point, I'm all ears.

I also don't mind taking out the entire point a).

barbieri: I'd take it out... what I see in practice is that thing "sticks" and then it's hard to remove the bad karma :-)

All parts of the system may have bugs, in that case they will be reported and fixed.

barbieri:
+1 for me, very good.

doc/howtos/building-images.rst
pohly: Resolved by taking it out in PR #100 in commit da1b663c60.

pohly:
PR #100 is currently more recent than this PR here. I suggest we merge PR #100 first, then rebase this one here and merge it later.

doc/howtos/building-images.rst
gvancuts: +1 on adding a subset of the README.md file to our main doc (via .rst file).

pohly:
Rebased on top of the recent app-framework image feature change and also fixed that (must be added to OSTRO_IMAGE_PKG_FEATURES). It also turned out that "full" is a reserved bundle name (with some subtle implications, like include os-core in pack-full-from-0.tar when it should have contained only the additional files), so now ostro-image-swupd.bb uses "world" as bundle name for "all our additional software".

PR #105, opened 2016-04-13
ostro-image: Include nodejs runtime tools in full images

nagineni:
Include nodejs runtime tools in full images

Signed-off-by: Sudarsana Nagineni <sudarsana.nagineni@intel.com>

### DIFF
--- a/meta-ostro/classes/ostro-image.bbclass
+++ b/meta-ostro/classes/ostro-image.bbclass
@@ -139,6 +139,7 @@ BUNDLE_CONTENTS[full] = " \
     ${FEATURE_PACKAGES_iotivity} \
     ${FEATURE_PACKAGES_java-jdk} \
     ${FEATURE_PACKAGES_node-runtime} \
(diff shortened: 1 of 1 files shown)


mythi:
+1

Let's wait for the build results.

pohly:
This needs to be rebased on top of PR #104. BUNDLE_CONTENTS[full] will no longer be in ostro-image.bbclass and the bundle had to be renamed to "world" in ostro-image-swupd.bb.

Actually, I think PR #104 already includes the nodejs-runtime-tools automatically, because now the content of the "world" bundle is derived directly from OSTRO_IMAGE_PKG_FEATURES, and that already has nodejs-runtime-tools. Previous the bundle was defined manually via a second, redundant list.

mythi:
@nagineni I've verified this indeed gets addressed by #104 so closing this.

nagineni:
@pohly @mythi Thanks for the update.

PR #107, opened 2016-04-14
ostro-os-production.inc: disable consoles/getty and mute logging

mythi:
This commit adds an image feature called 'muted' that can be
used disable all consoles and login services (getty).

In addition, 'muted' modifies systemd journald's log level such that
only emergency logs are logged. This essentially disables all debug
logs.

This leaves kernel printk setting unchanged. That is a BSP setting
Ostro does not need to touch.

The primary use for 'muted' is an a production image. Therefore, it
is added as a default there.

### DIFF
--- a/meta-ostro/classes/ostro-image.bbclass
+++ b/meta-ostro/classes/ostro-image.bbclass
@@ -61,6 +61,7 @@ OSTRO_IMAGE_PKG_FEATURES = " \
 # swupd = install swupd client and enabled generation of swupd bundles
 IMAGE_FEATURES[validitems] += " \
     autologin \
+    muted \
     ima \
     smack \
     swupd \
@@ -394,6 +395,39 @@ inherit xattr-images
 
 # Create all users and groups normally created only at runtime already at build time.
 inherit systemd-sysusers
+
+# Provide an image feature that disables all consoles and makes
+# journald mute.
+python() {
+    if bb.utils.contains('IMAGE_FEATURES', 'muted', True, False, d):
+        import re
+
+        # Mangle cmdline to drop all consoles and add quiet option
+        cmdline = d.getVar('APPEND', True)
+        new=re.sub('console=\w+(,\w*)*\s','', cmdline) + ' quiet'
+        d.setVar('APPEND', new)
+}
+
+# In addition to cmdline mangling, make sure journal only gets
+# emergency errors and any lower priority messages do not get
+# logged. This is temporary approach and the cleanest way to disable
+# logging (compared with removing the journald binary and related
+# services here). Once the systemd recipe gives us better granulatiry
+# for packaging, the preferred way to avoid logging is
+# to not install systemd-journald at all.
+ostro_image_muted () {
+    sed -i -e 's/^#\(MaxLevelStore=\).*/\1emerg/'\
+           -e 's/^\(ForwardToSyslog=yes\)/#\1/' \
+           ${IMAGE_ROOTFS}${sysconfdir}/systemd/journald.conf
+
+    # Remove systemd-getty-generator to avoid (serial-)getty services
+    # being created for kernel detected consoles.
+    rm ${IMAGE_ROOTFS}${systemd_unitdir}/system-generators/systemd-getty-generator
+
+    # systemd installs getty@tty1.service by default so remove it too
+    rm -r ${IMAGE_ROOTFS}${sysconfdir}/systemd/system/getty.target.wants
+}
+ROOTFS_POSTPROCESS_COMMAND += "${@bb.utils.contains('IMAGE_FEATURES', 'muted', 'ostro_image_muted;', '', d)}"
 
 # Disable images that are unbuildable, with an explanation why.
 # Attempts to build disabled images will show that explanation.
--- a/meta-ostro/conf/distro/include/ostro-os-production.inc
+++ b/meta-ostro/conf/distro/include/ostro-os-production.inc
@@ -2,5 +2,8 @@
 # in local.conf. In contrast to ostro-os-development.inc, no fallback
 # is provided.
 
+# Disable consoles and make journald quiet (=disable system logging).
+OSTRO_IMAGE_EXTRA_FEATURES_append = " muted"
+
 # Everything is (hopefully) ready, avoid initial sanity check.
 OSTRO_IMAGE_BUILD_MODE_SELECTED = "1"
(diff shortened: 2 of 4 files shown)


meta-ostro/classes/ostro-image.bbclass
```diff
@@ -381,6 +382,39 @@ inherit xattr-images
...
+# for packaging, the preferred way to avoid logging is
+# to not install systemd-journald at all.
+muted () {
+    sed -i -e 's/^#\(MaxLevelStore=\).*/\1emerg/'\
+           -e 's/^\(ForwardToSyslog=yes\)/#\1/' \
+           ${IMAGE_ROOTFS}${sysconfdir}/systemd/journald.conf
+
+    # Remove systemd-getty-generator to avoid (serial-)getty services
+    # being created for kernel detected consoles.
+    rm ${IMAGE_ROOTFS}${systemd_unitdir}/system-generators/systemd-getty-generator
```
pohly: I have a separate commit for this, something that we might include instead. See here:
https://github.com/pohly/ostro-os/commit/ac16ff65f5fab818b9f8e9b0331fc3a611598fa6

It relies on a OE-core patch which has been merged in the meantime.

meta-ostro/classes/ostro-image.bbclass
```diff
@@ -381,6 +382,39 @@ inherit xattr-images
 # Create all users and groups normally created only at runtime already at build time.
 inherit systemd-sysusers
 
+# Provide an image feature that disables all consoles and makes
+# journald mute.
+python() {
+    if bb.utils.contains('IMAGE_FEATURES', 'muted', True, False, d):
+        import re
+
+	# Mangle cmdline to drop all consoles and add quiet option
```
pohly: nitpick: wrong indention

meta-ostro/conf/distro/include/ostro-os-production.inc
```diff
@@ -2,5 +2,13 @@
 # in local.conf. In contrast to ostro-os-development.inc, no fallback
 # is provided.
 
+# Prefer systemd way of creating getty symlinks (instead of
+# the default systemd-serialgetty that creates everything in
+# do_install).
+PACKAGECONFIG_append_pn-systemd = "serial-getty-generator"
```
pohly: Ah, you are using the same systemd change as my patch.

But I think it is better to have it applied distro-wide, to minimize changes between images.

mythi: Makes sense. I did not want to have it too intrusive at this point but I'm OK to adopt

meta-ostro/classes/ostro-image.bbclass
```diff
@@ -381,6 +382,39 @@ inherit xattr-images
...
+}
+
+# In addition to cmdline mangling, make sure journal only gets
+# emergency errors and any lower priority messages do not get
+# logged. This is temporary approach and the cleanest way to disable
+# logging (compared with removing the journald binary and related
+# services here). Once the systemd recipe gives us better granulatiry
+# for packaging, the preferred way to avoid logging is
+# to not install systemd-journald at all.
+muted () {
```
pohly: I'd use "ostro_image_muted" here. Makes it easier to grep for it and avoids namespace issues (should someone else decide to add something called "muted").

Some of our other definitions in ostro-image.bbclass set a bad precedent here by using longer names but without consistent prefix.

meta-ostro/classes/ostro-image.bbclass
mythi: I'll fix it. I saw it also here after I had submitted the PR.

meta-ostro/classes/ostro-image.bbclass
barbieri: why not do it just in the kcmdline? https://www.freedesktop.org/software/systemd/man/systemd-journald.service.html says you can use `systemd.journald.forward_to_syslog=0` and `systemd.log_level=emerg` (but this one needs to be checked, may not be used by journald).

barbieri:
overall looks good. I'd consider this to imply in `quiet` being added to kcmdline as it does make sense. Maybe the `journal.conf` doesn't need to be modified if what I said about kcmdline can be used: `systemd.journald.forward_to_syslog=0` and `systemd.log_level=emerg`

meta-ostro/classes/ostro-image.bbclass
mythi: I thought about cmdline at some point but never tried it. log_level seems to be about systemd verbosity and not so much about what ends up in journal overall. However, I'll test it to be sure.

'''forward_to_syslog=0''' isn't necessary for muting. It gets enabled in systemd.bb and I'm just disabling it back.

mythi:
addressed the feedback, rebased, and force pushed.

@barbieri that cmdline approach did not work (journal was still full of debug) so I kept my original approach. 'quiet' was already in the original PR.

@pohly I ended up using systemd_%.bbappend since we already had in in meta-ostro. Looks reasonable?

kad:
I wonder how swupd would behave with such image hacks (sed files + removing something).
Anyone run swupd verify on resulted image ?

mythi:
swupd steps in after rootfs is finished. AFAICS we use ```ROOTFS_POSTPROCESS_COMMAND``` is used for many things for ostro.

kad:
it's big question when data for bundles collected. before or after this script part. if after, it means that swupd data would contain different state of the file. if after, it might be case where same file would be different between different bundles. Better to verify what actually end-up in manifests.

pohly:
On Fri, 2016-04-15 at 13:25 -0700, Alexander D. Kanevskiy wrote:
> it's big question when data for bundles collected. before or after
> this script part. if after, it means that swupd data would contain
> different state of the file. if after, it might be case where same
> file would be different between different bundles. Better to verify
> what actually end-up in manifests.

meta-swupd integration is intentionally designed so that the effect of
ROOTFS_POSTPROCESS_COMMAND influences also the bundles. It's the main
reason why we are not taking shortcuts, like just taking files from
packages: because then, further post-processing would not work in
combination with swupd.

This can be tested of course in this case, but I am confident that it'll
just work.

mythi:
I rebased the PR, split ostro.conf ```PNBLACKLIST``` to its own commit and force pushed an update.

Any +1?

mythi:
test this please

https://github.com/solettaproject/soletta/issues/1897 triggered on Edison.

mythi:
test this please

pohly:
Looks okay to be merged.

PR #110, opened 2016-04-15
Swupd doc

igor-stoppa:


doc/architecture/software-update.rst
```diff
@@ -35,7 +63,7 @@ and consumed by the software update client running on supported devices.
 
 Update information is secured by generating hashes of each file and then
 signing the file containing the references to the hashes.
-A secure channel is not required to distribute the updates.
+There is no need for a secure channel, for distributing the updates.
```
pohly: The opposite is true at the moment: because the swupd client does no checking of the data that it receives, secure updates are only possible when using https and relying on securing the transport channel.

If the document is meant to describe the ultimate goal, then there should at least be a remark that explains the current state.

dbkinder: Are you saying that the swupd client is NOT using hash and signature checks to validate the received data?  Do we need to remove this paragraph, or what would we replace it with?

pohly: swupd client is indeed not using hash and signature checks to validate the received data. What Igor describes here is the long term goal, not the current implementation.

I don't know what the document is meant to describe (both?), so I don't have a specific proposal for a replacement. Given that people will read this document to understand how Ostro OS currently works, at least the current status should be mentioned.

igor-stoppa: Let's say that I was too optimistic about the feature being available on time.
I will amend the doc.

doc/architecture/software-update.rst
```diff
@@ -21,11 +21,39 @@ Ostro OS based software can be deployed to a target device in two ways:
   component.
 
 
+Software Update
+===============
+The main concept behind the Software Update tool_ is that it doesn't
```
gvancuts: Could we make the link name to be 'Software Update tool' instead of just 'tool' (use backticks for to enclose the complete text in this case)

doc/howtos/software-update-server.rst
```diff
@@ -0,0 +1,46 @@
...
+Software Update: Building an Ostro |trade| OS  Repository
+#########################################################
+
+This technical note provides instructions to help you build a repository 
+used to make software updates work with the Ostro OS.
+
+Prerequisites
+=============
+
+- additional (fast) disk space: the tool is provide das additional layer, therefore it doesn't
```
gvancuts: s/provide das/provided as

gvancuts:
Not sure how to comment on the existing content of the file (that isn't being touched by this PR) but here are a few things I'd like that we modify:
  This is a component that Ostro OS borrows from Clear Linux\* OS
  for Intel |reg| Architecture and consists of both a server and client
  component.
This may give the impression that the concept of SW updates and bundles is only applicable to Intel-based platforms. Unless this is the case, I would suggest that we only refer to Clear Linux OS (drop the 'for Intel Architecture' in the sentence.

  It is similar to the concept of a package used in the vast majority of linux distros,
  but in general one software update bundle maps to several packages (ex: .rpm or .deb)
At first, this made me believe that we still use .rpm or .deb. I'd prefer if we said something like 'but in general one software update bundle is a collection of software components that used to be available as separate packages'

In general, can we add a few words about how this mechanism deals with dependencies? Is there such concept of bundles depending on other bundles?

igor-stoppa:
@gvancuts 
* ok for dropping the IA part
* wrt packages, well, we do use kinda of packages - this is how bitbake populates the rootfs. They are not .deb nor .rpm, but I still do consider them as packages.
* bundles do not depend on each other. If 2 bundles share some files, they list said files as dependency inside their respective manifests, but that's all there is to say about it. Dependency are handled only at package level when building the rootfs of each bundle.

gvancuts:
Hi @igor-stoppa , I realise we still use packages but this is what I consider part of the 'plumbing' now, it's not really visible to users, nor is it useable at all once you have an image, it's really only used by the Yocto build system. I was trying to find a better phrasing to indicate that bundles are a collection of SW components that would, in a traditional system, be packages... maybe @dbkinder could come up with something even better than my proposal.

IMHO, adding a, even small, paragraph explaining what you just say about dependencies may help clarify this for those (like me) who wondered how you'd deal with that. Your description makes sense although I do not yet fully comprehend the implications of this. E.g. what happens when you want to remove a bundle (when we'll support that) and it has files in common with other bundle(s)? It also sounds like in theory you could look at the list of bundles (ls /usr/share/clear/bundles) and effectively have some installed but not listed there (e.g. if a bundle includes another one completely)?

igor-stoppa:
On 19 April 2016 at 17:51, Geoffroy Van Cutsem <notifications@github.com>
wrote:

> Hi @igor-stoppa <https://github.com/igor-stoppa> , I realise we still use
> packages but this is what I consider part of the 'plumbing' now, it's not
> really visible to users, nor is it useable at all once you have an image,
> it's really only used by the Yocto build system.
>
​Well, not so much. If you want to create own layers, there is no way
around this. You are back to packages / package groups.

The bundles are really just a method for delivery, but they do not
completely hide the underlying methods used for populating a rootfs.

> I was trying to find a better phrasing to indicate that bundles are a
> collection of SW components that would, in a traditional system, be
> packages... maybe @dbkinder <https://github.com/dbkinder> could come up
> with something even better than my proposal.
>

​But that's what I think is misleading. The closest thing to bundles imho
is to think about a meta package that actually does contain what would be
pulled in from the dependency.​
​However I'm intentionally avoiding this because one of the key difference
between bundles and packages is that packages can have conflicts​ that will
surface only at installation time.
With bundles, bugs aside, every possible conflict is solved during the
creation of the bundle, which is an offline operation, from the perspective
of the target device.

> IMHO, adding a, even small, paragraph explaining what you just say about
> dependencies may help clarify this for those (like me) who wondered how
> you'd deal with that. Your description makes sense although I do not yet
> fully comprehend the implications of this. E.g. what happens when you want
> to remove a bundle (when we'll support that) and it has files in common
> with other bundle(s)?
>
​Nothing, imagine that each file is refcounting how many bundles need it.
Removal happens only when reaching 0 (files belonging to the core bundle
never do because the core bundle is unremovable.)

Here's a link to the presentation I gave at ELC:

http://events.linuxfoundation.org/sites/events/files/slides/-OpenIoT%20Summit%20-%20IoT%20devices-%20secure%20boot%20and%20sw%20maintenance_0.pdf

Does it seem any better?
Anything you would like to have lifted from there?

> It also sounds like in theory you could look at the list of bundles (ls
> /usr/share/clear/bundles) and effectively have some installed but not
> listed there (e.g. if a bundle includes another one completely)?
>

​I do not know what happens in that case.
I would expect the swupd-server to be smart enough and prevent the creation
of bundles where one is a subset of another​.
This needs to be tested.


-- 
igor

igor-stoppa:
On 19 April 2016 at 18:21, Stoppa, Igor <igor.stoppa@intel.com> wrote:

> Well, not so much. If you want to create own layers,

arghh s/layers/bundles/

-- 
igor

pohly:
On Tue, 2016-04-19 at 08:22 -0700, Igor Stoppa wrote:
> > It also sounds like in theory you could look at the list of bundles
> (ls
> > /usr/share/clear/bundles) and effectively have some installed but
> not
> > listed there (e.g. if a bundle includes another one completely)?
> >
> 
> ​I do not know what happens in that case.
> I would expect the swupd-server to be smart enough and prevent the
> creation of bundles where one is a subset of another​.

I think such a situation is fine.

Suppose foo completely contains bar.

Then you can have "foo" installed and you will (indirectly) also have
everything from bar installed, without having "bar" listed as installed
bundle.

The difference compared to also installing "bar" is that future updates
will ensure that everything from "bar" remains available, which becomes
relevant should at some point "bar" contains something that is not in
"foo".

igor-stoppa:
> I was trying to find a better phrasing to indicate that bundles are a
> collection of SW components that would, in a traditional system, be
> packages... maybe @dbkinder <https://github.com/dbkinder> could come up
> with something even better than my proposal.
> 

[igor]​But that's what I think is misleading. The closest thing to bundles imho
is to think about a meta package that actually does contain what would be
pulled in from the dependency.​
​However I'm intentionally avoiding this because one of the key difference
between bundles and packages is that packages can have conflicts​ that will
surface only at installation time.
With bundles, bugs aside, every possible conflict is solved during the
creation of the bundle, which is an offline operation, from the perspective
of the target device.






> IMHO, adding a, even small, paragraph explaining what you just say about
> dependencies may help clarify this for those (like me) who wondered how
> you'd deal with that. Your description makes sense although I do not yet
> fully comprehend the implications of this. E.g. what happens when you want
> to remove a bundle (when we'll support that) and it has files in common
> with other bundle(s)?

[igor]
​Nothing, imagine that each file is refcounting how many bundles need it.
Removal happens only when reaching 0 (files belonging to the core bundle
never do because the core bundle is unremovable.)

Here's a link to the presentation I gave at ELC:

http://events.linuxfoundation.org/sites/events/files/slides/-OpenIoT%20Summit%20-%20IoT%20devices-%20secure%20boot%20and%20sw%20maintenance_0.pdf

Does it seem any better?
Anything you would like to have lifted from there?

> It also sounds like in theory you could look at the list of bundles (ls
> /usr/share/clear/bundles) and effectively have some installed but not
> listed there (e.g. if a bundle includes another one completely)?
>

​I do not know what happens in that case.
I would expect the swupd-server to be smart enough and prevent the creation
of bundles where one is a subset of another​.
This needs to be tested.

igor-stoppa:
That might well be a way to fake pseudo-dependency, by generating Matroska-Bundles, but it really seems like a corner case.

dbkinder:
I'm not sure of the status of this PR and what changes are planned.  As it sits, I think folks are asking for changes before the merge is accepted but I'm not sure what those changes are from the discussion above, and when they'll be made?

mythi:
@dbkinder there was some feedback @igor-stoppa said he'd fix:
- add a note that client does not check for file hashes
- clarifiy it's not IA specific

After these, there were some typos still to be fixed. Those can be handled later when polishing the doc.

gvancuts:
I'm all good on the discussion, thanks for the extensive feedback @igor-stoppa and @pohly .

@igor-stoppa , the presentation you gave at ELC is very good and I think it deserves to be publicized from the main Ostro website. I'll talk to the web team about that. I also think it's a good reference to be added to our doc directly if you care to do that.

Thanks,
Geoffroy

barbieri:
Note that while we discuss this hash/signature, they are almost done with
the code that does that. So if not bother changing it.

Em quinta-feira, 21 de abril de 2016, Mikko Ylinen <notifications@github.com>
escreveu:

> @dbkinder <https://github.com/dbkinder> there was some feedback
> @igor-stoppa <https://github.com/igor-stoppa> said he'd fix:
>
>    - add a note that client does not check for file hashes
>    - clarifiy it's not IA specific
>
> After these, there were some typos still to be fixed. Those can be handled
> later when polishing the doc.
>
> —
> <https://github.com/ostroproject/meta-ostro/pull/110#issuecomment-212739250>
>


-- 
Gustavo Sverzut Barbieri
Mobile: +55 (19) 99225-2202

igor-stoppa:
I didn't realize github would loose the connection between the fork I used to created this PR and a new form I made, even if I used the same commit id and branch name. So now I have to close this and create a new PR :-(

PR #113, opened 2016-04-18
busybox: disable klogd and syslogd

mythi:
The default busybox configuration builds klogd and syslogd. These
daemons also get started on the target.

Ostro uses systemd journal that captures the same logs so klogd
and syslogd are rendundant.

Drop the redundant daemons by adjusting the busybox configuration.

Fixes: https://github.com/ostroproject/ostro-os/issues/31

### DIFF
--- a/meta-ostro/recipes-core/busybox/busybox_%.bbappend
+++ b/meta-ostro/recipes-core/busybox/busybox_%.bbappend
@@ -1,0 +1,11 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+
+# ostro.cfg disables CONFIG_SYSLOGD so the corresponding packaging
+# needs to be dropped as well.
+SYSTEMD_PACKAGES = ""
+PACKAGES_remove = "${PN}-syslog"
+RRECOMMENDS_${PN}_remove = "${PN}-syslog"
+
+SRC_URI_append = "\
+    file://ostro.cfg \
+"
--- a/meta-ostro/recipes-core/busybox/files/ostro.cfg
+++ b/meta-ostro/recipes-core/busybox/files/ostro.cfg
@@ -1,0 +1,2 @@
+# CONFIG_KLOGD is not set
+# CONFIG_SYSLOGD is not set


mythi:
test this please

(just a gigabyte test failure but I'll re-trigger anyway)

mythi:
test this please

pohly:
Looks good to me, please merge once we have clean test results.

mythi:
rebased and force pushed to restart a build.

meta-ostro/recipes-core/busybox/busybox_%.bbappend
```diff
@@ -0,0 +1,11 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+
+# ostro.cfg disables CONFIG_SYSLOGD so the corresponding packaging
+# needs to be dropped as well.
+SYSTEMD_PACKAGES = ""
+PACKAGES_remove = "${PN}-syslog"
+RRECOMMENDS_${PN}_remove = "${PN}-syslog"
+
+SRC_URI_append = "\
+    file://ostro.cfg \
```
ipuustin: I would maybe have named this file `drop_logd.conf` or similar...

pohly:
On IRC it was pointed out that we now use "ar" from Busybox, which (according to RP) probably shouldn't be compiled at all.

I can't prepare a PR for that unless this one here gets merged first, because I would also do it via the same busybox_%.bbappend.

@mythi if you modify this PR once more (for example, by renaming ostro.conf -> drop_logd.conf, which seems like a valid point), can you add a "disable-ar.conf" directly in this PR?

mythi:
@pohly I can take care of ```ar``` too in this PR. However, I'm not sure about the naming. If we are making ostro specific customization why not keep it all ostro.conf. I understand the idea with independent config fragments but it does not fit with busybox as nice as with the kernel.

pohly:
@mythi if we use independent fragments, some derived user could selectively undo our changes with SRC_URI_remove = "file://drop_logd.conf" in a .bbappend, and it would be more obvious in our .bbappend what it is that we change.

But I don't feel strongly about this.

mythi:
obsoleted by #128

PR #118, opened 2016-04-19
swupd: enable updating of /etc

pohly:
/etc was already getting included in bundles (because meta-swupd patches and
reconfigures swupd-server), now we also enable that for the client. At the same
time we re-configure Ostro OS such that the only content in /etc in the rootfs
directories is read-only.

This is an intermediate step towards stateless. Image updates from
"some files in /etc" to "no files in /etc" should work, too, because
we'll simply stop putting files into /etc and then the swupd client
will remove them (assuming we keep "update /etc" enabled in the
client).

@rojkov @igor-stoppa please review.

Replaces https://github.com/ostroproject/ostro-os/pull/58

### DIFF
--- a/meta-ostro/conf/distro/include/stateless.inc
+++ b/meta-ostro/conf/distro/include/stateless.inc
@@ -15,3 +15,70 @@ STATELESS_ETC_WHITELIST += "*"
 STATELESS_ETC_DIR_WHITELIST += "*"
 
 ###########################################################################
+
+# As step towards full stateless Ostro OS, we now
+# treat some files in /etc as conceptually read-only (i.e. neither
+# modified by the OS at runtime nor by an admin). Anything contained
+# in the rootfs directories will get bundled and added or updated when
+# running "swupd update".
+#
+# The implication is that we must keep certain files out of the rootfs
+# which do get modified at runtime, because otherwise there are
+# "swupd verify" failures.
+
+# mtab needs to be a symlink to /proc/mounts, probably forever.
+# There is no point in patching that out of binaries, nor is there
+# a need to customize it, so the symlink can remain there as read-only
+# system component.
+STATELESS_ETC_WHITELIST += "mtab"
+
+# OE-core puts some files into /etc which systemd then later overwrites
+# unconditionally via /usr/lib/tmpfile.d/etc.conf or creates dynamically
+# (machine-id). Therefore we can remove the redundant files from our rootfs
+# by not packaging them in the first place.
+STATELESS_RM_pn-systemd += " \
+    resolv.conf \
+"
+
+# machine-id has to be present in images at least as an empty file
+# because we might boot with the rootfs read/only. Otherwise
+# creating it during early boot fails (see
+# systemd/src/core/machine-id-setup.c).
+#
+# We solve this by keeping it in the rootfs and only excluding it
+# from the bundles. Short-term solution, something like this will
+# have to be handled by meta-swupd.
+STATELESS_ETC_WHITELIST += "machine-id"
+# This postfunc runs after the do_copy_bundle_contents task and ensures that
+# machine-id is removed from the bundle directories (to be moved into meta-swupd,
+# probably using the STATEFUL_FILES variable).
+STATEFUL_FILES += "/etc/machine-id"
+python stateless_rm_config_files () {
+  for bundle in d.getVar('SWUPD_BUNDLES', True).split() + ['os-core']:
+    bundledir = d.expand('${DEPLOY_DIR_SWUPD}/image/${OS_VERSION}/') + bundle
+    for tgt in d.getVar('STATEFUL_FILES', True).split():
+      try:
+        # Can't use os.path.join() because of the leading slash in tgt.
+        path = bundledir + tgt
+        bb.note('stateless: trying to remove %s from bundle %s = %s' % (tgt, bundle, path))
+        os.remove(path)
+      except Exception as e:
+        bb.note('Failed to delete file: %s due to %s' % (tgt, str(e)))
+}
+do_swupd_update[prefuncs] += "stateless_rm_config_files"
+
+# Depend on the installed components and thus has to be computed on
+# the device. Handled by systemd during booting or updates.
+STATELESS_RM_ROOTFS += " \
+    udev/hwdb.bin \
+"
+
+# Disable creation of /etc/ld.so.cache in images and bundles. The file
+# gets already recreated by systemd anyway when booting. Has to be
+# done by unsetting LDCONFIGDEPEND (checked by rootfs.py, which
+# creates the ld.so.cache) for all Ostro OS images, but not the
+# ostro-initramfs, so we cannot set it unconditionally.
+python () {
+    if bb.data.inherits_class('ostro-image', d):
+        d.setVar('LDCONFIGDEPEND', '')
+}
--- a/meta-ostro/conf/distro/include/stateless.inc
+++ b/meta-ostro/conf/distro/include/stateless.inc
@@ -1,0 +1,17 @@
+INHERIT += "stateless"
+
+###########################################################################
+
+# Temporary overrides until Ostro OS is fully stateless.
+
+# This entry here allows everything in /etc because Ostro OS is
+# not actually stateless yet. We merely use the stateless.bbclass
+# to remove files from /etc which get written on the device
+# and thus must be excluded from images and, more importantly,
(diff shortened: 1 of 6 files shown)


mythi:
I've checked this and it looks +1 but I need another review from @rojkov  or @igor-stoppa. Perhaps @juimonen as well. Please check.

rojkov:
+1

meta-ostro/conf/distro/include/stateless.inc
```diff
@@ -0,0 +1,60 @@
...
+# swupd bundles.
+STATELESS_ETC_WHITELIST += "*"
+
+# Empty directories must be kept.
+STATELESS_ETC_DIR_WHITELIST += "*"
+
+###########################################################################
+
+# As an intermediate step towards full stateless Ostro OS, we now
+# treat some files in /etc as conceptually read-only (i.e. neither
```
igor-stoppa: Even when Ostro becomes really stateless, a user might want to add some component which is not.
Having facility in place that can help coping with the statefulness seems like a good idea, compared to requiring the user to fix the component.

igor-stoppa:
+1, the latest build seems to fail for unrelated reasons.

meta-ostro/conf/distro/include/stateless.inc
gvancuts: In meta-ostro/conf/distro/include/stateless.inc<https://github.com/ostroproject/meta-ostro/pull/118#discussion_r60541662>:

> +# Temporary overrides until Ostro OS is fully stateless.

> +

> +# This entry here allows everything in /etc because Ostro OS is

> +# not actually stateless yet. We merely use the stateless.bbclass

> +# to remove files from /etc which get written on the device

> +# and thus must be excluded from images and, more importantly,

> +# swupd bundles.

> +STATELESS_ETC_WHITELIST += "*"

> +

> +# Empty directories must be kept.

> +STATELESS_ETC_DIR_WHITELIST += "*"

> +

> +###########################################################################

> +

> +# As an intermediate step towards full stateless Ostro OS, we now

> +# treat some files in /etc as conceptually read-only (i.e. neither

Even when Ostro becomes really stateless, a user might want to add some component which is not.
Having facility in place that can help coping with the statefulness seems like a good idea, compared to requiring the user to fix the component.

+1

mythi:
test this please

pohly:
I have not been able to reproduce the /var/log error, and Alexandru said that QA also ran into it in full tests outside of this PR.

I have enhanced the test in https://github.com/01org/meta-intel-iot-security/pull/116

I suggest we get that into ostro-os first, then re-test this PR here.

pohly:
retest this please

pohly:
retest this please

pohly:
Force-pushed a revised 9f4e4c6: it turned out that /etc/machine-id is needed in the images, so some kind of solution which allows installing it without also including it in bundles is needed. I'm now using something as a short-term solution (updating the intermediate image file lists) while @incandescant works on something similar for meta-swupd.

@incandescant: I tried the solution you have in https://github.com/incandescant/meta-myhouse/blob/master/classes/myhouse-core-qa.bbclass#L6 but I still saw the files appear in "full" and "os-core" manifests (the ones in the final www directory). I'm also not sure how copying the files into swupd images with empty do_rootfs would have worked.

Please have a look at my current solution and let me know whether this is okay as a short-term solution for Ostro.

barbieri:
systemd ships a service that creates that file, which should be per-board and can't be pre-generated

pohly:
On Mon, 2016-04-25 at 13:28 -0700, Gustavo Sverzut Barbieri wrote:
> systemd ships a service that creates that file, which should be
> per-board and can't be pre-generated

The solution is to ship an empty /etc/machine-id in the image and not
include that empty file in the bundles. Then systemd can overwrite the
empty file at runtime without causing problems during "swupd verify" or
"swupd update".

pohly:
retest this please

Failed to flash gigabyte_1 (https://ostroproject.org/jenkins/view/Test/job/test_gigabyte/827/console)

incandescant:
Apologies, my proposed solution didn't work as I forgot that when using os.path() 

> "If a component is an absolute path, all previous components are thrown away and joining continues from the absolute path component."

Our approaches attempt to achieve the same goal different ways; that is ensuring the file isn't copied into the bundle directory in www/image/

This workaround seems OK to me

> I'm also not sure how copying the files into swupd images with empty do_rootfs would have worked.

I'm not sure what you mean?

pohly:
@incandescant my comment was about ensuring that /etc/machine-id really gets copied into all images, include the  derived ones which have an empty do_rootfs. Your approach works, mine didn't, so I've force-pushed another revision of the final patch.

barbieri:
On Mon, Apr 25, 2016 at 10:31 PM, Patrick Ohly <notifications@github.com>
wrote:

> On Mon, 2016-04-25 at 13:28 -0700, Gustavo Sverzut Barbieri wrote:
> > systemd ships a service that creates that file, which should be
> > per-board and can't be pre-generated
>
> The solution is to ship an empty /etc/machine-id in the image and not
> include that empty file in the bundles. Then systemd can overwrite the
> empty file at runtime without causing problems during "swupd verify" or
> "swupd update".
>


I'm missing why we need to provide an empty /etc/machine-id. Isn't the idea
to not ship it (ie: not in rootfs OR in swupd), then it's completely "user"
owned (in this case, auto-generated)

pohly:
On Tue, 2016-04-26 at 07:13 -0700, Gustavo Sverzut Barbieri wrote:
> I'm missing why we need to provide an empty /etc/machine-id.

Because systemd does not boot correctly when the file is missing and
rootfs gets mounted read-only initially (as we do in Ostro). When that
happens, systemd also fails to create the missing file:

https://github.com/systemd/systemd/blob/master/src/core/machine-id-setup.c#L229

"System cannot boot: Missing /etc/machine-id and /etc is mounted
read-only.\n"
"Booting up is supported only when:\n"
"1) /etc/machine-id exists and is populated.\n"
"2) /etc/machine-id exists and is empty.\n"
"3) /etc/machine-id is missing and /etc is writable.\n");

barbieri:
can't we change the boot to rw?

On Tue, Apr 26, 2016 at 11:33 AM, Patrick Ohly <notifications@github.com>
wrote:

> On Tue, 2016-04-26 at 07:13 -0700, Gustavo Sverzut Barbieri wrote:
> > I'm missing why we need to provide an empty /etc/machine-id.
>
> Because systemd does not boot correctly when the file is missing and
> rootfs gets mounted read-only initially (as we do in Ostro). When that
> happens, systemd also fails to create the missing file:
>
>
> https://github.com/systemd/systemd/blob/master/src/core/machine-id-setup.c#L229
>
> "System cannot boot: Missing /etc/machine-id and /etc is mounted
> read-only.\n"
> "Booting up is supported only when:\n"
> "1) /etc/machine-id exists and is populated.\n"
> "2) /etc/machine-id exists and is empty.\n"
> "3) /etc/machine-id is missing and /etc is writable.\n");
>
>
>
> —
> You are receiving this because you commented.
> <https://github.com/ostroproject/meta-ostro/pull/118#issuecomment-214765274>
>



-- 
Gustavo Sverzut Barbieri
Mobile: +55 (19) 99225-2202

pohly:
@barbieri we could mount rw, but I find it preferable to solve the problem also for read-only rootfs, just in case that this becomes required at some point.

barbieri:
my fear is that empty `/etc/machine-id` results in temporary machine-id being generated and discarded at every boot. This would be bad since we need a persistent way to identify devices, like used for soletta APIs, one can ask http and others to include (optionally!) X-Machine-ID in requests so servers can identify devices.

pohly:
@barbieri don't worry, persistent machine-id works once we get past the boot issues.

pohly:
@mythi I think this is ready for merging.

Getting /etc updated is not release critical. OTOH I don't know what'll happen in terms of image updates when we turn it on later, so I have a strong preference to get this in before swupd starts getting used.

incandescant:
Right, that makes sense. iirc the derived bundle images use the manifest to determine their contents *after* do_swupd_update has run, so modifying the manifest during do_swupd_update results in the removed files being absent in the derived image.

8597b55 LGTM.

(Apologies, wrote this hours ago and forgot to hit 'Comment')

pohly:
Force-pushed rebased commits.

mythi:
The previous build was good and many +1's including me earlier so merging.

PR #121, opened 2016-04-19
ostro-image.bbclass: revert "Ostro: Start shell on init failure."

pohly:
This reverts commit 774953a0dec3e0227b243a14c16997ad82f1bef8.

We do not want debug options enabled unconditionally. In this case,
the debug option has not been needed for a while, so it is easier
to remove the feature entirely.


mythi:
test this please

PR #122, opened 2016-04-20
DOC: spelling and format fixes

dbkinder:
Fixed some spelling errors and some edits for clarity and formatting

[skip ci]

### DIFF
--- a/doc/architecture/system-and-security-architecture.rst
+++ b/doc/architecture/system-and-security-architecture.rst
@@ -33,8 +33,8 @@ full disk image is fast. Less common configurations are still
 possible, but may require compiling from source.
 
 The expectation is that most devices will have just a few trusted
-applications and not have access to an application store. Compared to,
-say, a mobile phone OS, the security focus is moved from protecting
+applications and won't have access to an application store. Compared to,
+for example, a mobile phone OS, the security focus is moved from protecting
 against malicious application to protecting against attacks coming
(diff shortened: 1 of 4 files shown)


doc/architecture/system-and-security-architecture.rst
```diff
@@ -417,20 +417,20 @@ For a detailed discussion of firewall configuration, see
 Production and Development Images
 =================================
 
-By default, building an image results in something that is locked-down
-and secure. This is how real products should be built. Unless some
-kind of application gets installed during image creation, one cannot
-do much with the running image (no user interface, no way to log into
-the system).
+By default, real product images should be built locked-down
```
pohly: I think this change looses one important part of the message, which is that compiling Ostro OS results in secure images by default. I.e. don't change the configuration and you are safe.

The proposed wording ("product images should be built locked-down") makes it sound either like we are not sure whether they really are locked-down or that this merely a recommendation.

pohly:
Looks good except for the part about "should be built locked-down".

doc/architecture/system-and-security-architecture.rst
dbkinder: OK.  Thanks for the comment.  Sometimes editing to make something read better can, like this, subtly alter the meaning.  I'll try again...

dbkinder:
I fixed the "should be built locked-down" ...

PR #123, opened 2016-04-20
ostro.conf: disable static libraries

pohly:
Poky switched to disabling static libraries in 2.1. Ostro OS should do
the same, see commit for the reasoning.

I have not actually tested the change. This will have to be done as
part of CI and QA testing.


kad:
👍

mythi:
@pohly thanks! Looks good. I'll add retest-pending label to run a fresh check once builds are good again.

mythi:
test this please

PR #124, opened 2016-04-20
doc: firewall configuration for services not run as root.

ipuustin:
Document the case where a service needs to be run with non-root permissions but the firewall commands within `ExecStartPre=` and `ExecStopPost=` options should be run as root.

### DIFF
--- a/doc/howtos/firewall-configuration.rst
+++ b/doc/howtos/firewall-configuration.rst
@@ -119,3 +119,19 @@ Opening holes to firewall using this method is not absolutely mandatory.
 If you are building a device without dynamic services or applications,
 you can simply add the necessary rules to the custom ruleset that is
 loaded upon device startup.
+
+Note that if you are starting your service as non-root user (as you
+probably should be) using systemd's ``User=`` option, you'll need to
+make sure that the firewall manipulation commands are run with original
+permissions. An easy way to do that is to use
+``PermissionsStartOnly=true`` option in the service file. This works if
+you don't need to run any other ``ExecStartPre=`` or similar commands
+with the reduced non-root privileges.
+
+Opening firewall holes for applications
+=======================================
+
+All this work for opening necessary firewall holes is abstracted away
+for applications. It's enough for applications to declare the ports that
+they are using in the application manifest. The application framework
+will take care of generating and applying suitable firewall rules.


ipuustin:
[skip ci]

doc/howtos/firewall-configuration.rst
```diff
@@ -119,3 +119,11 @@ Opening holes to firewall using this method is not absolutely mandatory.
...
 you can simply add the necessary rules to the custom ruleset that is
 loaded upon device startup.
+
+Note that if you are starting your service as non-root user (as you
+probably should be) using systemd's ``User=`` option, you'll need to
+make sure that the firewall manipulation commands are run with original
+permissions. An easy way to do that is to use
+``PermissionsStartOnly=true`` option in the service file. This works if
+you don't need to run any other ``ExecStartPre=`` or similar commands
+with the reduced non-root privileges.
```
gvancuts: Where would you put such line? In the service file itself or the *.conf file? I'm asking because I'm also wondering how that works in the case your service file is auto-generated by the AppFW.

ipuustin: In either the service (if it is a new Ostro OS -specific service that is being created) or in the conf file if the service file already exists. The conf files are concatenated to the service files by systemd.

gvancuts: Would that require a manual edit of the file generated by the AppFW?

barbieri: Em quarta-feira, 20 de abril de 2016, Geoffroy Van Cutsem <
notifications@github.com> escreveu:

> In doc/howtos/firewall-configuration.rst
> <https://github.com/ostroproject/meta-ostro/pull/124#discussion_r60389378>
> :
>
> > @@ -119,3 +119,11 @@ Opening holes to firewall using this method is not absolutely mandatory.
> >  If you are building a device without dynamic services or applications,
> >  you can simply add the necessary rules to the custom ruleset that is
> >  loaded upon device startup.
> > +
> > +Note that if you are starting your service as non-root user (as you
> > +probably should be) using systemd's ``User=`` option, you'll need to
> > +make sure that the firewall manipulation commands are run with original
> > +permissions. An easy way to do that is to use
> > +``PermissionsStartOnly=true`` option in the service file. This works if
> > +you don't need to run any other ``ExecStartPre=`` or similar commands
> > +with the reduced non-root privileges.
>
> Would that require a manual edit of the file generated by the AppFW?
>
Of course not. This will be done by the generator



> —
> <https://github.com/ostroproject/meta-ostro/pull/124/files/16a0329042cef2237b9560fbf4538b6e9d36c92e#r60389378>
>


-- 
Gustavo Sverzut Barbieri
Mobile: +55 (19) 99225-2202

ipuustin: Yes, they are generated, but with a twist...

AppFW has to actually do this in a more complex way due to the fact that the `ExecStartPre=` command and similar declared in manifest have to be run with the same permissions as the app itself. Thus the firewall commands, which need to be run as root, have to be started from a different sub-service that the actual application service depends on. @klihub already has this working.

gvancuts: Having that all done by the AppFW generator is definitely great! Can we reflect the current state of affair in this paragraph? It could be as simple as stating that nothing is to be done if using the AppFW generator and a manifest. If that's the case, we should also enhance the AppFW documentation which only mentions nspawn as the container type that is supported and I don't see how to define the ports that the app needs in the manifest.

mythi:
@ipuustin thoughts on @gvancuts's feedback? I think it'd make sense to clarify what all this (firewall config) means to an appfw user.

ipuustin:
I added a paragraph for the application fw case. Application fw documentation can include the instructions of declaring ports within the overall application manifest description.

doc/howtos/firewall-configuration.rst
```diff
@@ -119,3 +119,19 @@ Opening holes to firewall using this method is not absolutely mandatory.
...
+
+Opening firewall holes for applications
+=======================================
+
+All this work for opening necessary firewall holes is abstracted away
+for applications. It's enough for applications to declare the ports that
+they are using in the application manifest. The application framework
```
gvancuts: Can you include a reference to the Application Framework document here (hint: you can use :ref:`application-framework` for that)

mythi:
I'm merging this as it is now and let's handle the last comment from @gvancuts in a separate PR. That's because updates to this causes CI to build everything (some problem with skip ci)

PR #125, opened 2016-04-20
ostro-image.bbclass: temporarily disable soletta-dev-app

mythi:
The npm install used in the .bb is causing troubles in CI build
environment and it makes many pull request to fail due to network
issues.

Temporarily disable soletta-dev-app from the image to get integration
moving faster.


mythi:
@ipuustin FYI

@pohly is this enough or did I miss anything?

pohly:
I think removing "soletta-tools" in that one place should be enough.

mythi:
edit: added ```EXCLUDE_FROM_WORLD_pn-soletta-dev-app = "1"``` too

PR #128, opened 2016-04-21
busybox: disable components

pohly:
Replaces PR #113 and simplifies busybox even further.

Removal of "ar" and "netcat" is based on customer feedback.


mythi:
thanks! +1

PR #129, opened 2016-04-21
systemd-sysusers.bbclass: set login shell to /sbin/nologin

pohly:
systemd specifies that all users have sysusers will end up having
/sbin/login, which was missing from the helper class, leading to
unintentionally creating the users with /bin/sh instead.

@ipuustin that fixes the problem you found. Please review.

### DIFF
--- a/meta-ostro/classes/systemd-sysusers.bbclass
+++ b/meta-ostro/classes/systemd-sysusers.bbclass
@@ -23,7 +23,7 @@ systemd_sysusers_create () {
                     fi
                     comment=$(echo "$remaining" | cut -d '"' -f 2)
                     home=$(echo "$remaining" | cut -d '"' -f 3 | sed -e 's/^ *//' -e 's/ *$//')
-                    perform_useradd "${IMAGE_ROOTFS}" "$opts $uid --home-dir ${home:-/} --comment \"$comment\" $name" 10
+                    perform_useradd "${IMAGE_ROOTFS}" "$opts $uid --home-dir ${home:-/} --shell /sbin/nologin --comment \"$comment\" $name" 10
                     ;;
                   "")
                     ;;


ipuustin:
`/sbin/nologin`, you mean.

meta-ostro/classes/systemd-sysusers.bbclass
```diff
@@ -23,7 +23,7 @@ systemd_sysusers_create () {
                     fi
                     comment=$(echo "$remaining" | cut -d '"' -f 2)
                     home=$(echo "$remaining" | cut -d '"' -f 3 | sed -e 's/^ *//' -e 's/ *$//')
-                    perform_useradd "${IMAGE_ROOTFS}" "$opts $uid --home-dir ${home:-/} --comment \"$comment\" $name" 10
+                    perform_useradd "${IMAGE_ROOTFS}" "$opts $uid --home-dir ${home:-/} --shell /sbin/nologin --comment \"$comment\" $name" 10
```
ipuustin: The command looks right. Just change `/sbin/login` -> `/sbin/nologin` in the commit msg, and this is good to go.

pohly:
[skip ci]

pohly:
I force-pushed a rebased commit with updated commit message *after* adding the "skip ci" comment, but force-pushing seems to have triggered a new test build anyway?

@okartau should that have happened?

pohly:
@mythi @ipuustin okay now? The content is unchanged and the earlier test build was green, so IMHO this can be merged without another test. But that's now running, so we might as well wait.

mythi:
On Fri, Apr 22, 2016 at 11:08 AM, Patrick Ohly  wrote:

> I force-pushed a rebased commit with updated commit message *after*
> adding the "skip ci" comment, but force-pushing seems to have triggered a
> new test build anyway?
>
> @okartau <https://github.com/okartau> should that have happened?
>
>
There seems to be a problem with 'skip ci' added afterwards. #124 and #130
also get built.

ipuustin:
Looks good now!

PR #135, opened 2016-04-22
Ostro_security_flags.inc: removing not needed pkg exceptions

ereshetova:


mythi:
I've taken 9b9f123 in PR #142.  2fb210e we can keep until we see how soletta-dev-app plans to handle its modules.

Closing now.

PR #137, opened 2016-04-25
ostro-multilib.conf: Fix galileo build

Valtis:
This is a test PR.

Multilib build fails currently for ostro-image-swupd, as libgcc creates a symlink to the 32 bit glibc directory. As we use uclibc as the 32 bit library instead, this symlink is broken, and a sanity check fails during a swupd related task.

Suppress this symlink creation instead by removing 'lib32' from MULTILIB_VARIANTS-variable for libgcc, so that libgcc recipe does not realize we are using multilib feature.  

### DIFF
--- a/meta-ostro/conf/ostro-multilib.conf
+++ b/meta-ostro/conf/ostro-multilib.conf
@@ -24,3 +24,7 @@ EXCLUDE_FROM_WORLD_virtclass-multilib-lib32 = "1"
 # It would probably be nicer to do this as a linux-uclibc
 # override but that doesn't cover lib32-gcc-runtime
 SECURITY_CFLAGS_virtclass-multilib-lib32 = ""
+
+# libgcc produces a broken symlink during multilib_do_install to the non-existent 32 bit libgcc directory.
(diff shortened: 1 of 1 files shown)


pohly:
@Valtis is this test build still relevant? Can it be closed?

Valtis:
I asked QA to verify this fix, as testing the change itself in CI seems to be problematic (arm and corei7 targets aren't supported and the build fails). I'd keep it open for few days at least.

pohly:
@Valtis is this still considered just a test PR or do you want it to be merged now?

kad:
By labels, it is still test pr.

Valtis:
The problem with the PR is that I haven't found a way to have it pass the CI; by design it is intended only for Galileo and Edison. Both corei7-64 and arm builds fail. The current PR modifies the multilib conf file but it isn't active, so the pass here doesn't count. 

After discussion with @msivosuo, it was agreed that it should nevertheless be merged and then let the QA test/reopen the bug if there' still any issues.

kad:
If you have something machine specific, you can define variables based on target machine name.

PR #140, opened 2016-04-26
ostro.conf: disable SSTATE locked sigs checks

mythi:
The SDK_INHERIT_BLACKLISTed classes used in CI (and mirrored SSTATE)
cause locked signature file check failures with eSDK installation.

Until the blacklisted classes are fixed, SSTATE checks need to be
run in self-tests.


mythi:
@pohly could you take a look, thanks

pohly:
I have no idea what the problem was and what this change does, so I cannot really review it for correctness. If this has been discussed with someone (Yocto team, SDK owners?) they'll have to review.

If there's no-one else knows what this does either, then I suggest you just go ahead with merging it. At least the test results are okay.

mythi:
@pohly this was proposed by @bluelightning. 

taking it

PR #142, opened 2016-04-26
ostro.conf updates

mythi:
This PR removes some of the recipes from world build.

As we are now building the whole world for eSDK, we need to review
what are the recipes that are actually needed/useful.

With meta-openembedded layers (and some others too) we use
whitelist.bbclass to select which recipes are needed. However,
with openembedded-core, PNBLACKLIST is suited better.

This commit blacklists some recipes from openembedded-core that are
not applicable for Ostro.

Some further optimization is still needed. This should give a
safe start.


mythi:
v2: removed tempered as well since it depends on hid-api

mythi:
v3: rebased and force pushed with some updates

mythi:
v4: rebased, added 2bbf27c

mythi:
@pohly @ipuustin can you take a look please?

ipuustin:
Do you have a particular reason for not using `git revert`?

mythi:
@ipuustin no particular reason. it was quicker and less error prone this way since the revert conflict was messy.

ipuustin:
@mythi : Ok, that's a good enough reason. Otherwise the changes look good to me.

mythi:
v5: added 01df7ed: this helps making os-core leaner. The build was failing so added a new commit in this PR.

pohly:
+1 for merging, although I think that an explanation why some components are "not preferred" would be useful. I know that it is because of concerns about the upstream maintenance status, but others do not know that.

mythi:
v6: force pushed updates to commit messages based on feedback

PR #143, opened 2016-04-27
local.conf.sample: remove extra space in lines that can be uncommented

gvancuts:
Remove the extra space between '#' and the start of lines that can
be uncommented to tweak one's conf/local.conf file. This makes the
whole document more consistent as it is already done that way throughout
most of it and it also makes for a better user experience as forgetting
to remove the extra space at the start of the line generates a bitbake
parsing error.

[skip ci]


pohly:
+1, and I am saying that as the one who added most of these lines ;-)

I did so because I found it more readable when commented out, but indeed consistency and the "must also remove space" aspect are more important.

PR #144, opened 2016-04-27
python-pygobject: avoid gnome-common dependency

pohly:
gnome-common is not really needed (the native counterpart
pulled invia gnomebase.bbclass is enough). We can make
Ostro OS simpler and easier to review by not building it.


mythi:
nice, thanks for checking!

PR #150, opened 2016-04-29
meta-ostro: Required configuration changes for meta-java upstream change

Valtis:
ostro.conf, ostro_security_flags.inc and packagegroup-java-jdk.bb are updated to work with the new meta-java layer and different openjdk 8 recipe. ostro-x11.inc is removed as unnecessary.

Several recipes are added to the blacklist, as the removal of the X11 hack makes them unbuildable and recipe-selftest would fail otherwise. Second stage bootstrap vm is changed from Jamvm to Cacao vm for the same reason. This should also yield some performance benefits, as Jamvm is fully interpreter based, while Cacao vm uses JIT compiler. 


meta-ostro/conf/distro/ostro.conf
```diff
@@ -144,7 +144,7 @@ DISTRO_EXTRA_RRECOMMENDS += " ${OSTRO_DEFAULT_EXTRA_RRECOMMENDS}"
 QEMU_TARGETS ?= "arm i386 x86_64"
 
 # Set PREMIRRORS to prioritize Ostro OS autobuilder source cache.
-# The OSTRO_SOURCE_MIRROR_URL can be overriden in local.conf. 
+# The OSTRO_SOURCE_MIRROR_URL can be overriden in local.conf.
```
pohly: Please rebase on top of current master. While you are doing that, can you take out this unrelated whitespace change?

pohly:
Included in https://github.com/ostroproject/ostro-os/pull/85.

If that PR works, we can merge it and then this one here.

PR #152, opened 2016-04-29
ostro-image-swupd.bb: prepare for BUNDLE_FEATURES API change in meta-swupd

pohly:
meta-swupd was updated such that the API changed, causing
https://github.com/ostroproject/ostro-os/pull/60 to fail unless this
PR here gets merged into meta-ostro.

There's no big risk associated with it, but also no pressing need as
the updates of ostro-os are currently done manually and can exclude
meta-swupd.


barbieri:
+1

pohly:
Merging it because it is trivial and will remove one build failure.

PR #155, opened 2016-05-02
ostro.conf: Blaclkist BB_NUMBER_PARSE_THREADS for SDK

rojkov:
The variable BB_NUMBER_PARSE_THREADS set in CI in order to
limit the number of parse threads should not leak to the user's
environment since it may increase SDK installation time and
worsen runtime performance on capable systems.

Fixes: IOTOS-1548

### DIFF
--- a/meta-ostro/conf/distro/ostro.conf
+++ b/meta-ostro/conf/distro/ostro.conf
@@ -28,6 +28,12 @@ SDK_RECRDEP_TASKS = "do_deploy_files do_uefiapp"
 # Blacklist inherited classes that break proper data restore from SSTATE when populating the SDK.
 SDK_INHERIT_BLACKLIST = "buildhistory icecc buildhistory-extra buildstats-summary archiver isafw test-iot"
 
+# The variable BB_NUMBER_PARSE_THREADS set in CI in order to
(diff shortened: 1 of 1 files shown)


kad:
@okartau should we remove it from our settings ?

okartau:
Limiting of threads on CI worker was introduced to protect from overload, as worker, being a LXC container, sees all CPUs of phys host, which may be high. (some builders have 32 right now). BB uses num of CPUs as default num of threads. But we run multiple such LXC guests on top of same HW server  so we risk going out of balance. That at least is risk with build threads. For parse threads, there is likely no overload risk (?), so I think yes, we can try removing that setting from CI worker own setup.

pohly:
I suppose the change makes sense regardless whether the CI system has BB_NUMBER_PARSE_THREADS set, because other CI systems in which ostro.conf gets used may still have it. Agreed?

@okartau mentioned build threads. Is something similar needed for that?

Nitpick: Blaclkist spelling.

kad:
I'm ok with change itself, I'm just thinking that BB_NUMBER_PARSE_THREADS might be leftover that we can eliminate anyway.

rojkov:
Fixed the typo in the commit message.

PR #156, opened 2016-05-02
building-images: improvements

gvancuts:
Two commits to improve the building-images.rst file. One that fixes a typo and fixes some minor reST formatting issues (mainly double semi-colon that appeared in the final rendering). The second commit improves somewhat the section explaining how to add additional packages to an Ostro image by covering both noswupd and swupd cases.

[skip ci]

PR #158, opened 2016-05-02
DOC: Fix incorrect use of ::: for code blocks

dbkinder:
Scanned through all .rst files to change incorrect code block
ReST tag shorthand from ":::" to "::"  (settled on using "::" vs. ": ::" consistently)
and force pushed.

[skip ci]

### DIFF
--- a/doc/howtos/authorized-keys.rst
+++ b/doc/howtos/authorized-keys.rst
@@ -18,7 +18,7 @@ Generating a Private/Public Key Pair
 ====================================
 
 On Linux systems (or Windows systems with `Git Bash`_ installed) use the ``ssh-keygen`` utility to generate
-an ssh key pair by following its instructions ::
+an ssh key pair by following its instructions::
 
    $ ssh-keygen -t rsa -b 4096 -C "your_email@example.com"
 
@@ -60,7 +60,7 @@ Build-Time Configuration of Authorized Keys
 The recommended approach for enabling remote ``ssh`` access is to install authorized public ssh keys by 
 setting OSTRO_ROOT_AUTHORIZED_KEYS in your ``local.conf`` configuration file and building them into your image.
 
-#. Look in your ``local.conf`` file and find the line:::
+#. Look in your ``local.conf`` file and find the line::
 
    # OSTRO_ROOT_AUTHORIZED_KEYS = "ssh-rsa AAA...== john@example.com\nssh-dss AA...FPaQ== joan@example.com"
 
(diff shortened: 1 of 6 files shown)


pohly:
In PR #156 Geoffroy just just two colons directly after the last word. Wouldn't it be better to use that style consistently? Or are there reasons to use ": ::" instead?

gvancuts:
In PR #156<https://github.com/ostroproject/meta-ostro/pull/156> Geoffroy just just two colons directly after the last word. Wouldn't it be better to use that style consistently? Or are there reasons to use ": ::" instead?

As far as I know there is no difference but I would also prefer that we stay consistent and use "::". I opted for "::" earlier because that's already how it's done elsewhere in our documentation.

dbkinder:
I wanted the sentence preceding the block to have a ":" at the end indicating the block is a continuation of the thought, as in, use this command to do something interesting:
```
   $ interestingcommand --parameter=27
```
Alternatively, the preceding sentence should end with a period (i.e., ". ::") 

Looking through, for example, all the Yocto Project documents, I find an ending ":" is typically used and occasionally an ending period. We shouldn't end the sentence without any punctuation.

dbkinder:
I can go through all the documents to make this consistent, and document this in our documentation guidelines.  My vote is to have the preceding sentence before a block end with a colon (so using ": ::")

gvancuts:
The "::" right at the end of the entence will give you a single ':' ending it, it's a fully minimized version of ": ::" (although I agree the latter is more intuitive and user-friendly)


-------- Original message --------

dbkinder:
I didn't know about the difference between having and not having a space before the trailing "::" so I'll consider myself educated.  Thanks. Bottom line is that we want some punctuation at the end of these lines so I'll do a quick check.

dbkinder:
I scanned (again) for odd looking lines with :: on them and fixed and verified we don't have any more trailing lines in the generated documents with "::"

dbkinder:
@pohly it would be nice, but not critical to get this into the 1.0 documentation release

pohly:
@dbkinder The patch itself now is inconsistent: in authorized-keys.rst it uses ": ::" while in booting-and-installation.rst it uses "::".

Can we make this consistent across all documents?

My preference is for "::". It is shorter and more readable than the ": ::" where the single space is a rather subtle difference and does not "collapse" as easily into a single : when reading the .rst.

Regarding 1.0, we already have a release build, which is what is going to be tagged as 1.0. We could import all doc changes and tag that as 1.0_doc, then use that revision for the web update.

dbkinder:
@pohly @gvancuts all the ": ::" should be "::" now.

pohly:
@dbkinder no, I'm still seeing ": ::" in 23c9d483b, the current commit.  Can you check that you have force-pushed your revised commit?

pohly:
@dbkinder also please rebase.

pohly:
@dbkinder thanks, merging now.

dbkinder:
: ::)

pohly:
Argh 8 88)

PR #165, opened 2016-05-04
buildhistory-extra.bbclass: record ${S} and friends literally in history

pohly:
iotivity has
git://github.com/01org/tinycbor.git;protocol=https;name=cbor;destsuffix=${S}/extlibs/tinycbor/tinycbor
as entry in its SRC_URI. When we allow S to be expanded, we end up
with paths that vary from build to build.

Besides causing churn in the build history, it also causes iotivity to
show up once per architecture in the patch summary.

### DIFF
--- a/meta-ostro/classes/buildhistory-extra.bbclass
+++ b/meta-ostro/classes/buildhistory-extra.bbclass
@@ -28,10 +28,18 @@ python buildhistory_extra_emit_pkghistory() {
 
     import codecs
 
-    relpath = os.path.dirname(d.getVar('TOPDIR', True))
     pkghistdir = d.getVar('BUILDHISTORY_DIR_PACKAGE', True)
     if not os.path.exists(pkghistdir):
         bb.utils.mkdirhier(pkghistdir)
+
+    # Make the recorded information independent of varying paths.
+    # Some recipes use destsuffix=${S}/... to fetch components
+    # into their main source tree. The other variable do not show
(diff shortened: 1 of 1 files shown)


meta-ostro/classes/buildhistory-extra.bbclass
```diff
@@ -33,6 +33,15 @@ python buildhistory_extra_emit_pkghistory() {
     if not os.path.exists(pkghistdir):
         bb.utils.mkdirhier(pkghistdir)
 
+    # Make the recorded information independent of varying paths.
+    # Some recipes use destsuffix=${S}/... to fetch components
+    # into their main source tree. The other variable do not show
+    # up (yet), but might eventually.
+    d_original = d
```
kad: do we need to keep d_original ?

pohly: Not at the moment. I added it as a reminder that there are now two flavors of d, the simplified one and the original.

I don't care too much about it, though.

kad: better not to create dangling variables that assigned but never used.

pohly:
@kad removed the redundant d_original and also moved the initialized of "relpath" such that it uses the modified d variable, because the paths it gets compared against will also use the modified d.

pohly:
Now I pushed the right amended commit.

kad:
👍

PR #167, opened 2016-05-04
Enable minimal X11 baseline

jlaako:
Enable components to build and use minimal X11 baseline.

Minimal X11 environment can be included by using following line in
local.conf:
include conf/distro/include/ostro-os-core-x11.inc


kad:
looks ok to me.

meta-ostro/conf/distro/include/ostro-os-core-x11.inc
```diff
@@ -0,0 +1,3 @@
+DISTRO_FEATURES_append = " x11"
+DISTRO_FEATURES_append = " opengl"
+IMAGE_INSTALL_append = " packagegroup-core-x11 xterm twm"
```
barbieri: can we skip xterm and twm?

meta-ostro/conf/distro/ostro.conf
```diff
@@ -238,6 +238,9 @@ PNWHITELIST_openembedded-layer += " \
     libsocketcan \
     pkcs11-helper \
     vim \
+    libxaw \
+    xterm \
+    twm \
```
barbieri: would skip these 3, pure legacy

barbieri:
I'd keep xterm and twm out. Then we check the SDL approval for Gtk/xfce and enable that instead. What do you think?

kad:
xterm is real minimum to have, before we have something else.

barbieri:
Why? Just run from ssh with export DISPLAY=:0

Later if we add xfce we can add it's terminal, no need to have two.

kad:
just to be able to run some x11 example programs locally.

barbieri:
I'd not bother with that for now, as I said, xfce will going to be added later and it can bring xfce-terminal or xterm.

My rationale is that if we include something like ostro-crosswalk.inc, then we'd still be adding ostro-x11 and getting twm and xterm for no good reason.

pohly:
The proposed usage of the .inc file will put the "IMAGE_INSTALL" also into ostro-initramfs and thus include X11 in the initramfs. We have OSTRO_IMAGE_EXTRA_INSTALL for this, see https://github.com/ostroproject/meta-ostro/blob/master/doc/howtos/building-images.rst#installing-additional-packages

What exactly is the intended usage of the .inc file? Is it for real production devices or just for demo purposes? If it is "for real", then like Gustavo I'd prefer to have the package selection limited to just the essentials, i.e. packagegroup-core-x11.

kad:
at the moment mostly demo.

pohly:
If it's just demo, then the package selection doesn't matter. Perhaps add a comment that the package selection is just an example.

gvancuts:
+1 on switching from IMAGE_INSTALL to OSTRO_IMAGE_EXTRA_INSTALL

meta-ostro/conf/distro/include/ostro-os-core-x11.inc
jlaako: Default xinitrc uses those. Without those you don't get anything at all included that would have sensible output on X11. Many people like to run some tests on the image using terminal.

jlaako:
@pohly the intended users of the .inc file is the same as the people who would use the developer inc that gives root shell. Essentially it makes "startx" command useful on console.

jlaako:
I will make the IMAGE_INSTALL change.

jlaako:
New version, removes unnecessary cairo bbappend and changes the IMAGE_INSTALL

pohly:
Why do you need to extend the blacklisting? I would understand if you *removed* blacklist entries, but *adding* them?

Are the comment for these new entries really correct? For example, does distcc really depend on gtk+?

```
PNBLACKLIST[distcc] = "Cannot be built due to missing gtk+, not supported."
```

jlaako:
@pohly Yes, adding "x11" disto feature enables lot of gtk+ dependent features that need to be blacklisted. distcc being one of the many examples:

PACKAGECONFIG ??= "${@bb.utils.contains('DISTRO_FEATURES', 'x11', '${GTKCONFIG}', '', d)} popt"
PACKAGECONFIG[gtk] = "--with-gtk,--without-gtk --without-gnome,gtk+"

pohly:
So the blacklisting is a workaround for upstream making the assumption that gtk is enabled and desirable when "x11" is set as distro feature? And presumably this breaks the eSDK world-build, when adding x11?

This is all relevant information that should go into our repo, either in the commit message or the ostro.conf. Here in the pull request it will not be found in the future.

jlaako:
@pohly Yes... I don't know anything about eSDK world-build, so I cannot comment on that.

I can update the commit message to include comment about gtk+ assumption. ostro.conf is too messy to put something like that sensibly there (the entire black/white listing is an ugly mess).

pohly:
Yes, please update the commit message.

I agree that the black/white listing needs to be cleaned up, that's why I wanted to know exactly why something had to be added to it.

jlaako:
Commit message updated

jlaako:
Rebased and fixed merge conflict

pohly:
Okay, let's merge. The actual goal (building with x11 enabled and without building gtk) can be achieved differently, but that's something that can be fixed later.

PR #170, opened 2016-05-09
ostro_security_flags.inc: removing soletta exception for Wformat flags

ereshetova:
This exception is not needed since it complies fine with these flags

Signed-off-by: Elena Reshetova <elena.reshetova@intel.com>


ereshetova:
@pohly, I have created a clean pull request that is against tip of meta-ostro master

pohly:
Still compiles fine -> merging.

PR #171, opened 2016-05-09
iotivity-node: fix signed/unsigned char compiler problems

pohly:
Started to fail in recent CI pull, exact reason unknown. The code
clearly isn't handling signed char correctly, hence the patch.


pohly:
@gabrielschulhof: fix matches https://github.com/otcshare/iotivity-node/issues/49

Do you agree that Ostro OS should fix it locally for now, or can you prepare an update of meta-iot-web soon?

gabrielschulhof:
@pohly I can certainly fix this in iotivity-node, since we have some other things that need fixing because of iotivity's recent move from ```uint8_t``` to ```char```. @nagineni can then submit the package for integration. In the meantime, if it's fastest for Ostro, then, definitely, go ahead and use this patch until I release a fixed version.

PR #176, opened 2016-05-11
various distro meta data maintenance

pohly:
Bumping release number and bblayer.conf version.

### DIFF
--- a/meta-ostro/conf/distro/ostro.conf
+++ b/meta-ostro/conf/distro/ostro.conf
@@ -1,6 +1,12 @@
 DISTRO = "ostro"
 DISTRO_NAME = "Ostro OS"
-DISTRO_VERSION = "1.0~snapshot-${DATE}"
+# The version is bumped after each official release.
+# This implies that new releases go out with a version
+# string based on the predecessor plus the suffix, which
+# was deemed sufficient in the "[Ostro-dev] versioning Ostro OS"
+# mail thread.
+DISTRO_VERSION = "1.0+snapshot-${DATE}"
+DISTRO_VERSION[vardepsexclude] = "DATE"
 SDK_VENDOR = "-ostrosdk"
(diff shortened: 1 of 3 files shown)


pohly:
@kad okay to merge?

@Valtis all Beaglebone black devices were blacklisted during this PR's testing.

Valtis:
Beaglebone power cutter required cable unplugging again but it should work now.

pohly:
I don't think we should test this again; all we need is a +1 from someone.

meta-ostro/conf/distro/ostro.conf
```diff
@@ -1,6 +1,12 @@
 DISTRO = "ostro"
 DISTRO_NAME = "Ostro OS"
-DISTRO_VERSION = "1.0~snapshot-${DATE}"
+# The version is bumped after each official release.
+# This implies that new releases go out with a version
+# string based on the predecessor plus the suffix, which
+# was deemed sufficient in the "[Ostro-dev] versioning Ostro OS"
```
gvancuts: How about adding a link to it, i.e.: https://lists.ostroproject.org/pipermail/ostro-dev/2016-May/000427.html ?

pohly: I find that to long and distracting. If someone is interested, they can find the discussion in their favorite location (local copy, via a web search engine, gmane if he had registered our list, etc.)  via the mail thread title.

kad:
I would skip comments about mailing lists, but otherwise all ok.

pohly:
Seems there's no consensus about how much the comments should say about the background discussions ;-} Opinions range from "provide direct link" to "not mention it at all". Let me make a judgment call here and stick to the middle ground, which is what I have now - mentioning that there was a discussion (because we do not want to start a discussion from scratch the next time someone has questions about that line) but without the URL.

I'll merge now because it might fix build issues.

PR #177, opened 2016-05-11
ostro.conf: Remove packages from blacklist

rojkov:
Since cairo and x11 can be built in Ostro now there's no
need to blacklist packages that depend on them.

Also meta layers like meta-ros depend on the blacklisted
packages.


pohly:
I'm working on getting rid of the entire white/black listing. Let's wait with updating the current list until I'm done.

Or is there an urgent need to get this merged?

pohly:
This fails to build (https://ostroproject.org/jenkins/job/build-recipe-selftests/1646/console) because gtk+ now is enabled in our world build, but cannot actually be compiled due to the missing x11.

@rojkov can you check why gtk+ depends on virtual/libx11? gtk+.inc only adds that dependency when the x11 package feature is enabled, which normally isn't set when x11 is not a distro feature.

We recently had a similar problem with a meta-ostro change for cairo, see https://github.com/ostroproject/ostro-os/pull/90/commits/0dbed17dd27504338024e86728263aa101635db1

rojkov:
@pohly Yep, will check. I've just finished with meta-ros tweaks.

kad:
it comes from: meta-ostro/recipes-gnome/gtk+/gtk+_%.bbappend

pohly:
So basically another hack. I didn't find it because I was grepping for PACKAGECONFIG.

Most likely 02f9e6231acb1b can be removed. Please check all hacks added around that, Jussi only removed the one for cairo.

rojkov:
The file `meta-ostro/recipes-core/icedtea/icedtea7-native_%.bbappend` has been removed already at some point. What's left from that commit is only `meta-ostro/recipes-gnome/gtk+/gtk+_%.bbappend`.

I tried to include `packagegroup-java-jdk` to my image with `meta-ostro/recipes-gnome/gtk+/gtk+_%.bbappend` removed and it has built successfully. I'll update this PR to remove the bbappend.

PR #178, opened 2016-05-11
VirtualBox appliance generation: Change nic mode, delete temporary files and rename conflicting directory

Valtis:
VirtualBox uses a directory under /tmp for ipc. This can cause issues
if user already has VirtualBox installed, as the directory may already
exist and may be owned by root. The directory is renamed to avoid any
name collisions.

The VirtualBox working directory is also deleted, as this gets published, which is not intended.

Finally VirtualBox NIC is configured to use bridged mode instead of the default NAT, as this makes connecting to the VM easier from the host machine.

### DIFF
--- a/meta-ostro/classes/image-dsk.bbclass
+++ b/meta-ostro/classes/image-dsk.bbclass
@@ -67,7 +67,7 @@ create_ova() {
   STORAGE_DEVICE="0"
   STORAGE_DEVICE_TYPE="hdd"
 
-  
+
   RAW_IMAGE="${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.dsk"
   VIRTUALBOX_IMAGE="${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.vmdk"
   VIRTUALBOX_EXECUTABLE="${STAGING_BINDIR_NATIVE}/VBoxManage"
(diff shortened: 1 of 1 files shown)


Valtis:
@igor-stoppa Does this look good for you?

kad:
not good from my point of view. Two simultaneous builds on same host might mess with each other.
Why not to use mktemp and directory name pass via env variable or something like this ?

igor-stoppa:
yes, we just discussed this and agreed to have a unique id in the file name
(ex: bitbake pid)

On 11 May 2016 at 14:26, Alexander D. Kanevskiy <notifications@github.com>
wrote:

> not good from my point of view. Two simultaneous builds on same host might
> mess with each other.
> Why not to use mktemp and directory name pass via env variable or
> something like this ?
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/ostroproject/meta-ostro/pull/178#issuecomment-218432303>
>



-- 
igor

Valtis:
Round 2. It turns out you can pass the directory name through environmental variable, so I dropped the patch as long as we are OK that the directory resides under /tmp (that is hardcoded).

Also some whitespace fixes, as my editor ate some unnecessary whitespace. I noticed this only after pushing the changes. I'll fix it if it is an issue.

igor-stoppa:
Nitpicking: in the rm phase I would reuse the variable, if it's still available, instead of its string value.

igor-stoppa:
@Valtis: I am confused by the comment "It turns out you can pass the directory name through environmental variable" when then you remove a file. To which directory are you referring to?

Valtis:
Updated PR.

I mean the temporary directory VirtualBox creates under /tmp (/tmp/.vbox-......-ipc/files here)

igor-stoppa:
codewise LGTM  +1, spaces aside

for the spaces, I could close an eye, anyway this patch doesn't go anywhere

Valtis:
retest this please

Edison timed out during flashing.

Valtis:
Working directory is now deleted, as this gets published with the images. The appliance is also now configured to use bridged networking instead of using NAT, as this makes connecting to the VM easier.

igor-stoppa:
@Valtis looks good, maybe you could reword the title of the PR, to highlight that there are also these 2 new changes.

pohly:
Edison tests failed to run. Merging anyway.

PR #179, opened 2016-05-12
ostro.conf: Remove packages needed for XFCE from blacklist

rojkov:
XFCE might be a requirement for some Ostro-based distros.


pohly:
This causes libva to be included in our world build, and that then fails because virtual/mesa is not available.

The new recipe selection mechanism I am working should avoid this, but in the meantime you'll have to make the change depend on the x11 distro feature, as in:
```
PNBLACKLIST[vte] = "${@bb.util.contains('DISTRO_FEATURES', 'x11', '', 'Cannot be built due to missing gtk+, not supported.', d)}"
```

pohly:
Actually, perhaps just doing it for libva will be enough.

A quick test might be "bitbake --dry-run world" (untested).

barbieri:
mesa will be required soon to do the opengl, so I'd not bother much with that.

pohly:
The recipe selection without black/white listing is in a test PR now: https://github.com/ostroproject/ostro-os/pull/96

rojkov:
Added the condition and force pushed.

pohly:
retest this please

Some still ran with iotivity-node SRCREV 38f76790e2d0f4960b0f3360c7bbcdf5a69907f4. That's not what we have in ostro-os master now.

PR #180, opened 2016-05-13
update image content and configuration

pohly:
Adds ptest packages to ostro-image-swupd-dev, as discussed with QA.
@testkit please review

The SERIAL_CONSOLES fix is necessary to re-enable running of tests on
qemu. The dangling symlink was found while experimenting with dbg-pkgs
in addition to ptest-pkgs (unfortunately that makes the image content
too large).

Both get added here in this PR to save test cycles.

### DIFF
--- a/meta-ostro-fixes/meta-fixes/recipes-connectivity/openssl/openssl_%.bbappend
+++ b/meta-ostro-fixes/meta-fixes/recipes-connectivity/openssl/openssl_%.bbappend
@@ -1,0 +1,6 @@
+# Upstream-Status: Submitted [https://bugzilla.yoctoproject.org/show_bug.cgi?id=9523]
+
+do_install_ptest_append () {
+    # openssl.inc links to /lib/libcrypto.a instead of the correct /usr/lib/libcrypto.a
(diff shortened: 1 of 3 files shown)


pohly:
@kad, @igor-stoppa  what do you think about this?

meta-ostro-fixes/meta-fixes/recipes-connectivity/openssl/openssl_%.bbappend
```diff
@@ -0,0 +1,6 @@
+# Upstream-Status: Submitted [https://bugzilla.yoctoproject.org/show_bug.cgi?id=9523]
+
+do_install_ptest_append () {
+    # openssl.inc links to /lib/libcrypto.a instead of the correct /usr/lib/libcrypto.a
+    ln -sf ${libdir}/libcrypto.a ${D}${PTEST_PATH}
+}
```
kad: does it make sense at all ? I mean, static libraries going away, why to fix ?

kad:
generally ok...

pohly:
@kad the static libs are currently still there, so right now that problem still needs to be fixed.

PR #181, opened 2016-05-13
ostro images: introduce minimal image without GPLv3

pohly:
The minimal image is already useful without GPLv3, although achieving that goal at the same time would be nice.

### DIFF
--- a/meta-ostro/conf/distro/include/ostroproject-ci.inc
+++ b/meta-ostro/conf/distro/include/ostroproject-ci.inc
@@ -12,8 +12,20 @@
 # Generate summary statistics
 INHERIT += "buildstats-summary"
 
-# Enable CVE and other secutiry checks
+# Enable CVE and other security checks.
 INHERIT += "isafw"
+
+# Most of the images are expected to contain GPLv3
+# components. Therefore we only enable the license check for those
+# which must not have them (whitelisting), instead of excluding images
+# from the check (blacklisting).
+ISAFW_LA_PLUGIN_IMAGE_WHITELIST = "ostro-image-minimal ostro-initramfs"
+
+# ostro-image-minimal is only getting built below as proof that an Ostro OS root
+# filesystem is possible without depending on GPLv3 components. The actual
+# image files are not needed for that.
+IMAGE_FSTYPES_pn-ostro-image-minimal = ""
+OSTRO_VM_IMAGE_TYPES_pn-ostro-image-minimal = ""
 
 # Enable extended buildhistory:
 INHERIT += "buildhistory"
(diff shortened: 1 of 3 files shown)


pohly:
@dbkinder please review the patch.

Whether the minimal image works without GPLv3 remains to be seen and I'll update the commit message accordingly, but right now it looks like the actual change will be left unchanged.

dbkinder:
@pohly  What techwriter review is needed here? There are no documentation patches in this PR?

pohly:
@dbkinder: @kad wanted you to review this (see tag), I'm not so sure myself. I guess he wanted you to see this because it might have to be mirrored in our documentation.

kad:
Description about "magic" variables must be in documentation and not spread across dozens of config files comments or recipes. I've added tag because we need to consolidate those variables usage in some part of documentation.

pohly:
@kad: which "magic" variables do you mean? OSTRO_IMAGE_MINIMAL_EXTRA_FEATURES/INSTALL?

It is normal that a recipe has per-recipe variables. Do you expect all of them to be documented also outside of the recipe? That needs to be decided on a case-by-case basis. I think it would make sense in this case as part of documenting recommended usages and customizations of Ostro, but I believe we don't have that yet.

kad:
I expect if those variables are promoted for users to control how to install additional components into images, then those variables must be described in _documentation_ not spread over bunch of files in configs or recipes. That's why I added @dbkinder to review, if it is something to be described in e.g. "building-images" document.

pohly:
retest this please

gvancuts:
I'm starting to think that we should also have a glossary for all Ostro-specific variables. This would also point at http://www.yoctoproject.org/docs/2.1/ref-manual/ref-manual.html#ref-variables-glossary for all other Yocto generic variables.

@kad @pohly @dbkinder , thoughts?

kad:
I'm all for it. We need to have one document that would list variables that might control different aspects of the build and are specific to Ostro. Having descriptions spread across multiple include files, recipes, classes and even sometimes commit messages is not a good developer experience.

dbkinder:
Ah ha.  Understand.  Any thoughts on how to find all these configuration variables (grep for OSTRO in local.conf and the various other Yocto Project recipe and build files for example) and their purpose and then maintain the list moving forward if new ones are added -- I did check and the Yocto Project documentation folks and their variable list is maintained manually.

dbkinder:
in the meta-ostro repo, does this find all the OSTRO-specific variables we'd want to document (are they all in files within the meta-ostro folder of this repo?

grep -RIhio "\w*OSTRO\w*" meta-ostro | sort -f | uniq

gvancuts:
That command looks just fine to me. I'd run that from the ostro-os top-level folder just in case some are not in meta-ostro (in this case, I also exclude the build folder using --exclude-dir=build*).

I believe the ostro_image_* in small caps are internal function. They're not meant to be user-tuneable so I don't think we should include them in the glossary.

pohly:
retest this please

The GPLv3 image license check is in ostro-os master now. Once we have another test build we'll know whether the ostro-image-minimal is indeed free of GPLv3 components.

pohly:
The resulting ostro-image-minimal was indeed GPLv3-free, so this PR is now ready for merging. I have added a commit which configures the CI builds so that the isafw license check is only applied to the minimal image and the initramfs.

@dbkinder you reviewed the updated image comments and will include that in future documentation enhancements, right? Agreed to merge?

@testkit the license check is automated, but functional testing (basically just "it boots") currently has to be done manually (because we don't have a test configuration for the image). Therefore I am disabling the image creation for the minimal image, to save resources and avoid creating artifacts that we do not want to publish. Please confirm that you agree with this approach.

kad:
👍 from my side.

pohly:
isafw image license check is clean now.

PR #182, opened 2016-05-17
image-dsk.bbclass: Revert back to old network interface mode

Valtis:
When attempting to launch .ova that is set into bridged mode, it fails
with an error message. While quickly toggling the mode to nat and then
back to bridged fixes the issue, it still looks bad. As a short term
solution the interface is reverted back to NAT mode until I have time to
investigate what the root cause is.

### DIFF
--- a/meta-ostro/classes/image-dsk.bbclass
+++ b/meta-ostro/classes/image-dsk.bbclass
@@ -67,9 +67,6 @@ create_ova() {
   STORAGE_DEVICE="0"
   STORAGE_DEVICE_TYPE="hdd"
 
-  # Network adapter mode
-  NIC_MODE="bridged"
-
   RAW_IMAGE="${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.dsk"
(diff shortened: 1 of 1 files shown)


kad:
We never should ship with bridged by default in first place. it should be user's choice if he wants insecure image in his LAN.

pohly:
@kad when Ostro is installed on real devices, it also is directly connected to the local networks. So from that perspective it makes more sense to not use NAT.

kad:
if user install image on real device, this device most probably connected to some kind of "lab" network.
Desktop computer where virtualbox run, most probably connected to "office" network, which might be ok to have some testing stuff behind VirtualBox's NAT, but most probably not OK to have those directly bridged to office network. NAT is always safe choice.

pohly:
@kad our thread model and OS configuration do not rely on a "trusted" lab network. So using regular "office" network should be safe. I don't mind adding NAT as an additional defense, but arguing that it is because Ostro images are not safe to use in an office network is just not  a valid argument - if it was, we would have done something wrong ;-}

kad:
it's not about protecting ostro image from "office" network. it's more protecting "office" network from any random e.g. broadcasting service that might be part of image (dhcpd, mdns, etc).

PR #183, opened 2016-05-17
openjdk/jre-8: ignore warnings about deprecated functions

pohly:
readdir_r() recently got marked as deprecated, which causes builds
against OE-core master to fail. Work around it until meta-java
maintainers (notified) come up with something better.

This PR blocks further updates of OE-core in ostro-os (like
https://github.com/ostroproject/ostro-os/pull/100).


pohly:
@Valtis do you agree with merging this?

Valtis:
Looks good to me.

PR #184, opened 2016-05-18
ostro.conf: exclude gsignond from world build

pohly:
Building the eSDK indirectly builds gsignond and its plugins, which
fails when using latest libsoup from OE-core master:

../../git/src/gsignond-oauth-plugin.c: In function 'gsignond_oauth_plugin_init':
../../git/src/gsignond-oauth-plugin.c:540:5: error: 'soup_session_async_new_with_options' is deprecated: Use 'soup_session_new' instead [-Werror=deprecated-declarations]

This blocks https://github.com/ostroproject/ostro-os/pull/100, for example in 
https://ostroproject.org/jenkins/job/build_beaglebone/1613/console

@kad okay to merge without testing? Running a build test would not tell us
much when done with current ostro-os master.

[skip ci]


kad:
ok for me.

PR #186, opened 2016-05-18
replace black/white-listing

pohly:
The new approach is expected to be more flexible (= can be extended,
replaced or ignored by users of Ostro OS), easier to maintain (= a
single, complete list instead of adding and removing from the
default), and is meant to be used to enforce stricter checking of new
components.

Further enhancements to the supported-recipes.bbclass error message
are possible, but it should already be useful as it stands now.

@kad @elena this is the new mechanism that I was mentioning at some point.


meta-ostro/conf/distro/include/ostroproject-ci.inc
```diff
@@ -31,6 +31,9 @@ SDKMACHINE ?= "x86_64"
 # Generate source mirror tarballs:
 BB_GENERATE_MIRROR_TARBALLS = "1"
 
+# Do not allow building of anything that hasn't been checked beforehand.
+SUPPORTED_RECIPES_CHECK = "fatal"
+
 # Configure images in "development mode".
```
kad: I don't think it's good idea to put it here. This should be default value in distro config that user can override in local.conf if he/she wants.

kad:
In principle, I'm not against this. Something like this I was suggesting in the beginning, before whitelist was implemented. Back then it was two concerns:
- about maintaining such one big list, it was considered more problematic than whitelisting whole layer, e.g. oe-core.
- we might be interested in some recipe, but because we have multiple layers, we need to be specific in some cases where recipe comes from. e.g. there were issues where oic layer was shipping conflicting recipes and we were just whitelisting from that layer few recipes which we particularly were interested in.

So, to summarize: having check agains one big list of "approved" components is fine, but we need to be more careful on removing current whitelists, because as result we might got recipes from layers we don't want to.

gvancuts:
Could you also adjust whatever is needed in our documentation: howtos/building-images.html#whitelisting-a-recipe

@dbkinder , fyi

meta-ostro/conf/distro/include/ostroproject-ci.inc
pohly: I disagree. Consider what happens when we make SUPPORTED_RECIPES_CHECK = "fatal" the default: any custom recipe that a user of Ostro adds to his image now needs to be added explicitly to a custom list. That would be a very bad user experience, because most users won't care about controlling their build, at least not initially.

@gvancuts what do you think?

pohly:
@gvancuts I'll update the documentation as part of this PR, once we have addressed the initial review comments - see in particular the question whether the check should be enabled by default.

meta-ostro/conf/distro/include/ostroproject-ci.inc
kad: That's why I said about comment in local.conf. The less differences would be between CI builds and end-user builds - the less chances that SSTATE is not valid and less chances that we would see different errors than our users.
Another thing, does that feature support multiple lists files ? if so, then probably good idea to teach devtool to maintain own list for it's workspace as append to system-wide list.

pohly: Even enabling something that then most users will want to disable because it gets in their way would not be good user experience.

The way the mechanism works at the moment is such that it does not influence SSTATE, so while I understand the concern in principle, it does not apply here. It also does not influence the choice which version of a recipe gets built - I'll check that, but I suspect that we'll need to use blacklisting or layer priorities/ordering if we still need that kind of control.

Yes, multiple file lists are supported.

kad: we still need control on which recipes and from which layers got built, especially as we are dependant on so many upstream layers with conflicting recipes. We need to verify list of what we are getting as result with current master build before merging this PR

gvancuts:
@gvancuts<https://github.com/gvancuts> I'll update the documentation as part of this PR, once we have addressed the initial review comments - see in particular the question whether the check should be enabled by default.

Thanks!

pohly:
I found two cases where we end up building some different recipe than before:

* libmicrohttpd now comes from meta-oe because meta-soletta does not ensure that it has a higher priority (which it should have, because the intention of copying the recipe was (partly) to tweak it for soletta).
* xserver-nodm-init comes from meta-oe instead of oe-core - a name conflict which should be resolved in meta-oe (according to RP), but can also be solved with BBMASK

So bottom line is, in a perfect world we wouldn't need to check from which layer a recipe comes, but as we don't live in a perfect world, I need to extend  the check to also lock supported recipes to specific layers. Will do that...

pohly:
retest this please

AFT failures: https://ostroproject.org/jenkins/job/test_galileov2/1107/ https://ostroproject.org/jenkins/job/test_minnowboardmax/1107/

pohly:
@kad commented elsewhere that he would prefer to use the layer name instead of the collection name. However, layer names as shown in "bitbake-layers" or the repo revision dump in build output/build history are just the directory names and thus can be chosen by developers when checking out a layer repository.

The collection name is chosen by the layer maintainer and that is the only thing that identifies a recipe's origin reliably at the moment. @bluelightning confirmed that and suggested to stick with that until OE cleans this up by removing the "collection" concept. Also note that the current approach with black/white-listing also uses collection names, so that aspect is the same as before.

pohly:
retest this please

Another test rig failure (https://ostroproject.org/jenkins/job/test_gigabyte/1129/console).

kad:
What could be improved, even if we continue to use collection names, is the error reporting. Showing collection name doesn't help much, but showing error with actual vs. expected collection directory name would be more usable for user (in case he checkouts layers into another directories).

pohly:
@kad agreed, the error message can be improved. For example, it could list the current layer directory name of the new recipes, if that's possible to determine (not so sure). @bluelightning is that possible (even if it is a crude hack at the moment)? Do I have to write custom code (like "walk up the directory hierarchy until a conf/layer.conf is found")?

Adding the license, summary, and home page URL will also be useful for deciding whether the recipe can be supported.

As long as we keep using the "<recipe>@<collection>" as unique ID, the content of the error message can be copied into the supported recipe files.

However, I'm not exactly sure what you mean with "expected collection directory name". Where is that supposed to come from? Are you suggesting that we should add it to the supported recipes list? Using which format?

I'd prefer to enhance only the error message as much as possible, and keep the files as they are now.

kad:
By "expected collection name" I meant if we have in supported list recipe1@collection1 but in reality in build for some reason it selects recipe1@collection2, it should show in error that wrong version of recipe1 for some reason chosen and supported version is located at recipe1@collection1.

kad:
For devtool recipes, we need to have glob whitelisting.
Something like
```
*@workspacelayer
```
This would cover all custom user recipes that user would add on top of Ostro base.

pohly:
@kad: do we want globbing or regular expressions? Both can be done.

Like enhancing the error message this can still be added later without breaking the current code, because existing recipe names and collection names do not contain special characters.

So I'm still wondering whether we shouldn't simply merge this PR as it is and work on enhancements later - see also PR #188.

kad:
@pohly I don't have preference on glob vs. re.

As for merging as is, I'm not ok with that. While you're on it, fix it to the level that it would be usable now, not later.

pohly:
@gvancuts can you please comment on the developer usability aspect of this? Should the "supported recipes check" be enabled by default for anyone compiling Ostro OS from source (i.e. add it to the ostro.conf distro configuration, with instructions for disabling it via local.conf provided in the error message), or should it be enabled only in the CI (as done now)?

We can address some of my concerns (developers getting the error for their own recipes) by allowing all recipes in the local devtool workspace, but the error would still appear once developers move those recipes to a real layer.

I'd prefer to limit the check to our CI, but I don't mind changing it if you and/or Sasha want it enabled by default. Please discuss with Sasha.

kad:
@pohly we still need documentation for this new feature in either building-images document or on overall document that we discussed with @dbkinder, which describes Ostro specific flags and variables.

pohly:
@kad yes, I know that I need to revise the code and write documentation about it. But before writing the documentation I want to get closure on how the check will be enabled, hence my question to @gvancuts.

pohly:
Regarding glob vs. re, my plan is to go with re, because it is more versatile and emulating globbing is trivial (* = .*).

kad:
I think regardless of default mode of the check, documentation need to state what this feature used for, how to configure it, how to add custom lists and of course, format of those lists (with note about collection names).

RE sounds good for me. Do you plan to put extra check for those REs ? like, if they would apply separately on recipe name/collection name, or you plan to allow single RE that would cover both ?

pohly:
I'd prefer to write the documentation once, when it is clear how it is going to be used. More efficient that way.

My thinking regarding re is that it has two be two separate res, one for recipe name, one for collection. Otherwise one cannot do the "recipe foo is not supported in collection bar, only in collection(s) xyz" part that asked for (where xyz is a list of entries where the recipe part matches).

kad:
for RE: good. we are on same page.

pohly:
New code pushed with support for regular expressions and enhanced error message.

No change to the deployment (still limited to CI) and no docs yet, so still "author-rework-required".

meta-ostro/classes/supported-recipes.bbclass
```diff
@@ -0,0 +1,288 @@
...
+            if len(parts) != 2:
+                raise RuntimeException("%s.%d: entry must have format <recipe name regex>@<collection name regex>, splitting by @ found %d parts instead: %s" % (
+                    filename,
+                    linenumber,
+                    len(parts),
+                    pattern))
+            def parse_regex(regex):
+                try:
+                    # must match entire string, hence the '$'
+                    return (re.compile(regex + '$'), regex)
```
kad: if user for some reason uses $ in the end of his pattern, it shouldn't be added as last symbol.

meta-ostro/classes/supported-recipes.bbclass
```diff
@@ -0,0 +1,288 @@
...
+                    raise RuntimeError("%s.%d: parsing '%s' as regular expression failed: %s" % (
+                        filename,
+                        linenumber,
+                        regex,
+                        str(ex)))
+            self.pn_re = parse_regex(parts[0])
+            self.collection_re = parse_regex(parts[1])
+
+        def supported(self, pn, collection):
+            supported = bool((pn is None or self.pn_re[0].match(pn)) and (collection is None or self.collection_re[0].match(collection)))
```
kad: wrapping this line would help a lot in readability and further maintenance of this code

meta-ostro/classes/supported-recipes.bbclass
```diff
@@ -0,0 +1,288 @@
...
+}
+
+python supported_recipes_eventhandler() {
+    supported_recipes, files = load_supported_recipes(d)
+    supported_recipes_check = d.getVar('SUPPORTED_RECIPES_CHECK', True)
+    if not supported_recipes_check:
+        return
+
+    import re
+    isnative_exception = re.compile('|'.join((
```
kad: This seems to be good candidate to be configuration variable, as distro "minimal cross build tools"  might be different.

meta-ostro/classes/supported-recipes.bbclass
pohly: 'foobar$$' is a valid regex and identical to 'foobar$', therefore I didn't see a need to check for this special case. Do you still think the code needs to check for it?

meta-ostro/classes/supported-recipes.bbclass
pohly: okay

meta-ostro/classes/supported-recipes.bbclass
pohly: I'm not convinced this list has to be configurable because the content is mostly (exclusively?) determined by OE-core, but I can add a variable for it.

meta-ostro/classes/supported-recipes.bbclass
kad: I'm spoiled with RE libraries that not always behave the same way on different OS. in case of python on Linux, yeah, might be ok with double $.

meta-ostro/classes/supported-recipes.bbclass
kad: I just worried for scenarios where e.g. glibc would be replaced with musl. or uninative is cripled in.
Having those explicitly declared in visible place might help some developers in future.
E.g. if this list mandatory to OE-core, then put it as OE-core list.
Not critical to address now, just corner case that might surprise people who haven't looked at class implementation.

gvancuts:
Hi @pohly @kad , I've (finally) taken the time to review this thread. I'm thinking that enabling this by default is the right thing to do even it's it at the expense of user experience. We can mitigate this to some extent by clearly documenting this in our doc, have a bit text on that in the local.conf file *and* that the error message in case it gets on the way makes it very clear what is happening so people don't scratch their heads wondering what is happening.

Where could one maintain their own lit of packages? Could that be done initially in the local.conf directly and later move to a customer layer (and possibly combine both)?

Thanks

kad:
@gvancuts we minimally mitigate with whitelisting workspace that devtool uses.
So, for relative big portion of users, who are interested in few of their apps on top of Ostro it would be transparent. For people who would add their own layers, those would require to adjust lists or mode of the check.

pohly:
Okay, I'll update the integration so that the check is enabled and fatal by default and add the corresponding documentation. My goal is to merge this on Monday. As usual, doc enhancements by @dbkinder can follow in a later PR by him, once at least the initial documentation is available.

pohly:
One observation about the "add hint when recipe is supported elsewhere" commit.
Because the .*@workspacelayer entry matches all recipes, the resulting warning says:

 ERROR: The following unsupported recipes are required for the build:
   less@core (only supported in workspacelayer)

even when there is no "less" in the workspacelayer. There's no way how the code can check whether a recipe of that name exists in that other layer. Perhaps the error message should be changed to "would be supported", to make it clear that this is a hypothetical situation? Then it will be less confusing IMHO.

pohly:
I've include the proposed "would be supported in" change, replaced the check activation in the CI conf with the same change in ostro.conf, and updated the documentation.

I consider this ready for merging now, @kad @gvancuts please check.

kad:
IMHO, better not at all produce misleading warning if it can't be detected. I haven't verified line-by-line implementation how your code decides matching, but in theory, it should be possible to see if pattern is having wildcard characters and don't include them in suggestion hints.

Also, probably it make sense to move "+.*@workspacelayer" into own list file, so it would be demonstrating possibility of multi list. And also allow to replace list without need of replicating same line in derivative distros. Something like this:

```
SUPPORTED_RECIPES ?= "${META_OSTRO_BASE}/conf/distro/include/devtool-supported-recipes.txt ${META_OSTRO_BASE}/conf/distro/include/ostro-supported-recipes.txt"
```

BTW, we need to be careful with mode of check assigned in ostro.conf. Distro configuration file processed after local.conf and auto.conf, so it should have weakest assignment (??=) if we expect that user would be able to override it. Or it might be better to put default in local.conf.sample, instead of distro conf.

pohly:
Regarding the "would be supported/only supported in" hint: I had also considered checking whether the alternative is a regex or plain string, but discarded that idea because there could be cases where someone ends up using a regex like gcc-.*@core and wants to have the hint shown.

I no longer find the "would be supported in" hint misleading and I am reluctant to make intrusive changes at this point (at the very least, we need to update documentation and the line which adds the hint to the output, but then we have redundant code left). I suggest we keep it like this for now and revisit the topic after other people have also seen the message.

I've moved the @workspacelayer to devtool-supported-recipes.txt. It is indeed better that way.

The SUPPORTED_RECIPES_CHECK ?= in ostro.conf was already weak enough so that the user can override it in local.conf. However, I like the local.conf.sample approach better, so I have moved it there.

Testing failed with "org.jenkinsci.plugins.scriptsecurity.scripts.UnapprovedUsageException", let's see whether the updated branch has more luck.

kad:
retest this please

kad:
While you'll be fixing hardcoded list of toolchain packages (which I still believe  would be better to be explicitly listed in separate list file rather than in bbclass code), reverse please "solutions" in error message. First user need to understand dependencies, why is that. Than deal with lists and masks. And only as last resort disable check if not needed.

pohly:
Several new commits pushed, making the special cases configurable via variables, reordering the solutions, and finally addressing the "unsupported recipes" found during the selftest in the CI.

pohly:
retest this please

pohly:
retest this please

meta-ostro/conf/distro/ostro.conf
```diff
@@ -24,7 +24,9 @@ OSTRO_SDK_UPDATE_URL ??= "https://download.ostroproject.org/builds/ostro-os/late
 
 # Extensible SDK configuration: Ostro uses "minimal" type which minimizes the size of the downloaded
 # SDK itself and configures devtool to use SSTATE_MIRROR (defined in o sdk-extra.conf).
-# The package data available gets recorded.
+# The package data available gets recorded. This has the side-effect that exporting the eSDK
+# for a certain image will not only cover the image, but also all packages that ared included
```
gvancuts: typo: s/ared/are

pohly:
@kad, @gvancuts okay to merge? The previous test build succeeded except for beaglebone, where testing didn't start (https://ostroproject.org/jenkins/job/dsl-meta-ostro_pull-requests/1516/).

I fixed the typo found by @gvancuts, too.

gvancuts:
:+1: 

@dbkinder , let's do the editorial pass after it has been merged.

dbkinder:
Will do.

-- david kinder (Phone)

On May 30, 2016, at 7:22 AM, Geoffroy Van Cutsem <notifications@github.com<mailto:notifications@github.com>> wrote:


👍

@dbkinder<https://github.com/dbkinder> , let's do the editorial pass after it has been merged.

kad:
ok to merge once build/testing completes.

pohly:
Testing completed successfully, merging.

PR #188, opened 2016-05-23
Remove blacklisting for libva and related components

jlaako:
Since graphics is now supported, also remove blacklisting of libva and
related components to allow use of video acceleration API.

Fixes: IOTOS-1607


kad:
Should it be still controlled by distro features ? (e.g. x11)

pohly:
Agreed, this might have to be conditional, if it breaks world builds when x11 is not active. We'll see once testing runs, which somehow hasn't started yet.

test this please

pohly:
And indeed, it broke pretty quickly in the world build done by the sstate selftest.

Note that merging PR #186 would remove the need to maintain the backlist in such a complicated way. Ostro OS could keep these recipes out of the "supported recipes" list and a derived distro can add it to its own additional "supported recipes".

@kad what do you think about merging PR #186 as it is now and leaving the error message enhancements for later?

kad:
#186 I don't like to merge in current form right now.
Error message must be improved, then also I don't agree on mode of that checker.
#186 should not prevent other things to go first.

pohly:
PR #186 does not block merging black/white listing changes. What's blocking this PR right now is the build failure on Ostro OS, so I pointed out that there is an alternative to having @jlaako fix this PR.

jlaako:
With graphics enabled this builds fine for me. The recipes have pkgconfig checks for x11 && opengl distro features. See:
http://git.yoctoproject.org/cgit/cgit.cgi/meta-intel/tree/common/recipes-multimedia/libva

I don't quite get how I could make this conditional? Unless something is pulling those libva packages into images, there shouldn't be any problem having them non-blacklisted.

jlaako:
Due to various places where pkgconfig is used, trying to share sstate cache for x11/opengl and non-gfx distro features is likely futile?

jlaako:
Changed intel libva -driver depend on opengl distrofeature same way as libva, should work?

kad:
Yes, that looks better. Let's hope automatic testing wouldn't find something else.

msivosuo:
@pohly  - ready to merge?

PR #189, opened 2016-05-25
Add groupcheck to Ostro OS.

ipuustin:
Groupcheck is a drop-in polkit replacement for embedded systems. It only supports authentication by group membership. See https://github.com/ostroproject/groupcheck for details.

This PR adds groupcheck recipe and its default (empty) configuration. It also adds groupcheck to Ostro OS images.

### DIFF
--- a/meta-ostro/recipes-security/groupcheck-settings-default/groupcheck-settings-default_0.1.bb
+++ b/meta-ostro/recipes-security/groupcheck-settings-default/groupcheck-settings-default_0.1.bb
@@ -1,0 +1,28 @@
+# Copyright (C) 2016 Intel.
+# Released under the MIT license (see COPYING.MIT for the terms)
+
+DESCRIPTION = "Default groupcheck policy settings."
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+SRC_URI = " \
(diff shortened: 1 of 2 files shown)


barbieri:
+0

the package itself is okay, but I wonder if a feature `security` is the way to go. Maybe just add that by default given it's super-small overhead. 

also, let's add the docs for that.

ipuustin:
Yes, `security` is bit grandiose. :-) I was thinking there could later be more stuff in it... The docs need to concern mostly app fw. I'll talk to @klihub about that. Also what is missing from this PR is the policy, meaning the mapping of groups to allowed actions. That needs to be defined before groupcheck is useful in any way.

kad:
@ipuustin what do you think about splitting it into two PRs to discuss separately (recipe for groupcheck) and separately image configuration ? For recipe, no questions, all good from my side. For distro feature it require a bit more thinking.

pohly:
Let's remove the security feature. If we want to make the package optional, then we should let applications depend on it explicitly so that it is only installed when something expects it to be there. However, that begs the question what policy gets installed when that happens.

Long term, I'd prefer to have it enabled in all Ostro OS images, i.e. add it to the ostro-image.bbclass, together with a default configuration that implements a suitable set of rules. That then needs to be reflected in our architecture documents.

Short term I tend to agree with @kad: let's add the recipe but without enabling it in Ostro OS yet.

ipuustin:
Ok, sounds good. I'll remove the second patch from this PR.

PR #191, opened 2016-05-31
Software Update: add QA test Bundles into dev image

yma11:
Add 2 QA test bundles into dev image to cover verification of shared contents between bundles.
qa-bundle-a: contains 2 recipes, hello-bundle-a and hello-bundle-s(shared).
qa-bundle-b: contains 2 recipes, hello-bundle-b and hello-bundle-s(shared).

[Relates-to]: IOTOS-1290

Signed-off-by: yma11 <yan.ma@intel.com>

### DIFF
--- a/meta-ostro/recipes-image/images/ostro-image-swupd.bb
+++ b/meta-ostro/recipes-image/images/ostro-image-swupd.bb
@@ -112,8 +112,7 @@ SWUPD_IMAGES ?= " \
     all \
 "
 
-# In practice the same as "all" at the moment, but conceptually different
-# and thus defined separately.
+#add qa-bundle-a in swupd-dev image for QA testing.
 SWUPD_IMAGES[dev] = " \
     world-dev \
     qa-bundle-a \
--- a/meta-ostro/recipes-image/images/ostro-image-swupd.bb
+++ b/meta-ostro/recipes-image/images/ostro-image-swupd.bb
@@ -33,6 +33,8 @@ OSTRO_IMAGE_EXTRA_FEATURES += "swupd"
 # https://bugzilla.yoctoproject.org/show_bug.cgi?id=9493).
 SWUPD_BUNDLES ?= " \
     world-dev \
+    qa-bundle-a \
+    qa-bundle-b \
 "
 
 # os-core defined via additional image features maintained in ostro-image.bbclass.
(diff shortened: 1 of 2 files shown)


yma11:
Per previous discussion in mail thread, QA need to add 2 test bundles for verifying the behaviors of shared contents among bundles. @pohly @igor-stoppa , pls review this PR. Thanks.

yma11:
Hi @pohly the CI error is:
20:49:44 ERROR: The following unsupported recipes are required for the build:
20:49:44   hello-bundle-a@iotqa (would be supported in workspacelayer)
20:49:44   hello-bundle-b@iotqa (would be supported in workspacelayer)
20:49:44   hello-bundle-s@iotqa (would be supported in workspacelayer)
What should I do to resolve these errors?

pohly:
@yma11: Normally, you would have to add these entries to ostro-supported-recipes.txt in meta-ostro. However, in this case we can treat the bundles as a special case (helper recipe) which does not need to be listed explicitly. I'll make that change today.

meta-ostro/recipes-image/images/ostro-image-swupd.bb
```diff
@@ -105,7 +115,8 @@ SWUPD_IMAGES ?= " \
 # In practice the same as "all" at the moment, but conceptually different
 # and thus defined separately.
```
pohly: Please remove the comment. With your change, the "dev" image no longer has the same content as the "all" image.

pohly:
Ignore my earlier comment. I was misled by the naming of your hello-bundle-* recipes. Those are not bundles, they are normal recipes. I would have avoided using the "bundle" word here and instead used something like iotqa-example-recipe-a/b/c. But let's keep it as it is now.

So the correct solution is indeed what the error message says:
```
ERROR: The following unsupported recipes are required for the build:
  hello-bundle-a@iotqa (would be supported in workspacelayer)
  hello-bundle-b@iotqa (would be supported in workspacelayer)
  hello-bundle-s@iotqa (would be supported in workspacelayer)
...
To avoid this message, several options exist:
...
* Add the unsupported recipes to one of the following files:
  /work/meta-ostro/meta-ostro/conf/distro/include/devtool-supported-recipes.txt
 /work/meta-ostro/meta-ostro/conf/distro/include/ostro-supported-recipes.txt
...
```

In this case, it needs to go into ostro-supported-recipes.txt as part of your commit. Mind the alphabetic sorting.

pohly:
Not perfect (last commit mixes two different changes), but good enough. Merging.

PR #192, opened 2016-05-31
stay on gcc 5.3

pohly:
Some of our components (iotivity, Edison kernel) do not compile with
gcc 6.1 yet. Until they do, we need to stay on gcc 5.3.

This commit was tested successfully with the latest upstream
components in https://github.com/ostroproject/ostro-os/pull/110

@kad: agreed to use this as a workaround for the time being? Bugs against
iotivity and the kernel are filed, see comment in ostro.conf.


kad:
ok for time being.

PR #193, opened 2016-05-31
supported recipes: some enhancements

pohly:
Major new change is a report which also includes information about
source(s), home page and summary.

### DIFF
--- a/meta-ostro/classes/supported-recipes.bbclass
+++ b/meta-ostro/classes/supported-recipes.bbclass
@@ -270,8 +270,14 @@ python supported_recipes_eventhandler() {
             collections = supported_recipes.find_collections(pn)
             return ' (would be supported in %s)' % ' '.join(collections) if collections else ''
 
+        def re_escape(str):
+            # In contrast to re.escape(), we only escape + (as in gtk+3).
+            # Escaping all non-alphanumerics makes many entries (like linux-yocto)
+            # unnecessarily less readable (linux\-yocto).
+            return ''.join(['\\+' if x == '+' else x for x in str])
+
         logger('The following unsupported recipes are required for the build:\n  ',
-               '\n  '.join(sorted(['%s@%s%s' % (pn, collection, collection_hint(pn)) for pn, collection in unsupported.iteritems()])),
+               '\n  '.join(sorted(['%s@%s%s' % (re_escape(pn), re_escape(collection), collection_hint(pn)) for pn, collection in unsupported.iteritems()])),
                '''
 
 Each unsupported recipe is identified by the recipe name and the collection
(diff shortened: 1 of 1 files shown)


meta-ostro/classes/supported-recipes.bbclass
```diff
@@ -270,8 +364,14 @@ python supported_recipes_eventhandler() {
...
 
+        def re_escape(str):
+            # In contrast to re.escape(), we only escape + (as in gtk+3).
+            # Escaping all non-alphanumerics makes many entries (like linux-yocto)
+            # unnecessarily less readable (linux\-yocto).
+            return ''.join(['\\+' if x == '+' else x for x in str])
+
         logger('The following unsupported recipes are required for the build:\n  ',
-               '\n  '.join(sorted(['%s@%s%s' % (pn, collection, collection_hint(pn)) for pn, collection in unsupported.iteritems()])),
+               '\n  '.join(sorted(['%s@%s%s' % (re_escape(pn), re_escape(collection), collection_hint(pn)) for pn, collection in unsupported.iteritems()])),
```
kad: does collection name allow to have "+" inside ?

meta-ostro/classes/supported-recipes.bbclass
```diff
@@ -270,8 +364,14 @@ python supported_recipes_eventhandler() {
             collections = supported_recipes.find_collections(pn)
             return ' (would be supported in %s)' % ' '.join(collections) if collections else ''
 
+        def re_escape(str):
+            # In contrast to re.escape(), we only escape + (as in gtk+3).
+            # Escaping all non-alphanumerics makes many entries (like linux-yocto)
+            # unnecessarily less readable (linux\-yocto).
+            return ''.join(['\\+' if x == '+' else x for x in str])
+
```
kad: re.sub("\+","\\+", str) would work for your case instead of creating own function.
Also, "str" in python is builtin keyword. overriding it is not good.

meta-ostro/classes/supported-recipes.bbclass
pohly: I suppose it may happen, although I haven't seen it used in practice.

meta-ostro/classes/supported-recipes.bbclass
pohly: I can use re.sub(), but I would still do it in a utility function, to avoid cut-and-paste. Agreed, or would you prefer to have the same expression repeated, once for pn, once for collection?

kad: such small repetition is not that problematic. I think other function with only line of "beef" is usually not justified. In the end of the day, it's python, not java for all those bloated getters and setters :)

pohly:
@kad please re-review. I've inlined the re_escape(), using just regular str.replace() instead.

kad:
looks good.

PR #195, opened 2016-06-02
soletta_%.bbappend: adapt to OE-core packaging change

pohly:
This fixes the build error in our current CI PR (https://github.com/ostroproject/ostro-os/pull/116).

### DIFF
--- a/meta-ostro/recipes-soletta/soletta/soletta_%.bbappend
+++ b/meta-ostro/recipes-soletta/soletta/soletta_%.bbappend
@@ -4,6 +4,10 @@ DEPENDS_append = " systemd"
 
 SRC_URI += " file://config"
 
+# Restore "no strip, no debug split" behavior after
+# recent OE-core packaging change. See https://github.com/solettaproject/meta-soletta/issues/88
+INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
+
 INSANE_SKIP_${PN}-dev += "dev-elf"
 
 do_configure_prepend() {


kad:
why not directly to soletta layer ?

pohly:
@kad because I don't know whether the fix works with older OE-core releases (the test build will show that), and because I don't know how long it will take them to update meta-soletta. We can and should give them some time to fix it there. As plan B I prepared this PR here.

kad:
ok, but I think @brunobottazzini would be able to check that quickly.

barbieri:
I believe it was a left over workaround, we shouldn't force split or strip, be enabled or disabled.

brunobottazzini:
Indeed it is a left over. I will propose a PR for it in meta-soletta.

pohly:
Not sure whether it is related, but can you make sure that the revised soletta recipe passes the dev-elf QA test?

In Ostro, Mikko added a suppression for that in a soletta_%.bbappend:

INSANE_SKIP_${PN}-dev += "dev-elf"

```
commit 686fe481a0a13f09e1f90bbc319220db8e6c08ac
Author: Mikko Ylinen <mikko.ylinen@intel.com>
Date:   Thu Feb 4 10:21:47 2016 +0200

    soletta-dev: temporarily skip the new dev-elf QA check
    
    insane.bbclass recently added a new QA check to see all .so's
    in a -dev package are correctly pointing to the versioned .so.
    
    Soletta packaging does not satisfy this requirement and it
    requires bigger re-structuring to get this fixes.
    
    Therefore, temporarily skip the dev-elf QA check for
    soletta-dev.
    
```

It would be nice to get rid of that workaround, too.

barbieri:
indeed, it may require some work in the build system as it's not automake/libtool that does all for us. But we can get it done.

brunobottazzini:
@pohly, I started using Ostro and personally, I did not notice this QA problem. Maybe, because I haven't seen this workaround was in Ostro.

Someone started using meta-soletta without using the Ostro OS and, they opened an issue about that. I added the insane skip in the recipe to allow them to use it. 

So the problem is general with Ostro and Yocto.

I agree it would be nice get rid of this workaround.

pohly:
Work on fixing this in meta-soletta is in progress, but first attempt was not successful (https://github.com/solettaproject/meta-soletta/issues/88). Let's keep this PR open for a bit longer, in case that we need it, but not merge it yet.

pohly:
@brunobottazzini do you have an ETA for fixing the problem in meta-soletta? If it is still taking more (too much?) time, should we merge this PR here as a solution that works right now?

brunobottazzini:
@pohly you can merge this patch, I am focusing now in helping the 1.0 release of Soletta.

I didn't have much time to fix the problem.

pohly:
@kad can you please merge if you agree that we should take this route for addressing the problem?

kad:
For me this looks like workaround. Probably better to wait for proper fix, if there are no other reasons.

pohly:
@kad: so you propose to go back to manual merges which exclude OE-core? As @brunobottazzini just said, it is unclear when a proper fix will be forthcoming from meta-soletta. In the meantime importing OE-core updates is blocked.

kad:
As I  can see now in automatic PR, OE-core is problematic by itself right now :(

pohly:
Fixed via https://github.com/solettaproject/meta-soletta/pull/92

PR #196, opened 2016-06-05
buildhistory-extra: record full information of package

kad:
Put into buildhistory full information about binary package and
original recipe.

### DIFF
--- a/meta-ostro/classes/buildhistory-extra.bbclass
+++ b/meta-ostro/classes/buildhistory-extra.bbclass
@@ -120,21 +120,33 @@ python buildhistory_extra_emit_kernelconfig() {
 }
 
 buildhistory_get_image_installed_append() {
-	# Create a file mapping installed packages to recipes
-	printf "" > ${BUILDHISTORY_DIR_IMAGE}/installed-package-recipes.txt
-	cat ${IMAGE_MANIFEST} | while read pkg pkgarch version
-	do
-		if [ -n "$pkg" ] ; then
-			recipe=`oe-pkgdata-util -p ${PKGDATA_DIR} lookup-recipe $pkg`
-			pkge=`oe-pkgdata-util -p ${PKGDATA_DIR} read-value PKGE $pkg`
-			if [ "$pkge" != "" ] ; then
-				pkge="$pkge-"
-			fi
-			pkgr=`oe-pkgdata-util -p ${PKGDATA_DIR} read-value PKGR $pkg`
-			if [ "$pkgr" != "" ] ; then
-				pkgr="-$pkgr"
-			fi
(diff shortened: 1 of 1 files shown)


pohly:
Why are you adding LD_PRELOAD= ?

If you want to get rid if pseudo, then this is probably not sufficient (I was told that pseudo re-adds the variable, or something like that). Instead there's a pseudo-specific variable which disables pseudo in the new process.

From the description it is not clear what's the enhancement. Can you give a before/after example?

meta-ostro/classes/buildhistory-extra.bbclass
```diff
@@ -120,21 +120,31 @@ python buildhistory_extra_emit_kernelconfig() {
...
+            pkge=`LD_PRELOAD= oe-pkgdata-util -p ${PKGDATA_DIR} read-value PKGE $pkg`
+            if [ "${pkge}" != "" ] ; then
+                pkge="${pkge}:"
+            fi
+            pkgr=`LD_PRELOAD= oe-pkgdata-util -p ${PKGDATA_DIR} read-value PKGR $pkg`
+            if [ "${pkgr}" != "" ] ; then
+                pkgr="-${pkgr}"
+            fi
+            recipe=`LD_PRELOAD= oe-pkgdata-util -p ${PKGDATA_DIR} read-value PN $pkg`
+            recipev=`LD_PRELOAD= oe-pkgdata-util -p ${PKGDATA_DIR} read-value PE $pkg`
```
pohly: Should that be 'read-value PV'?

kad: Yes, indeed typo. :(

kad:
LD_PRELOAD, yes I wanted to see how it behave with a lot of forked processes, but looks like PSEUDO_UNLOAD is the way to go.

kad:
Problem: in our current buildhistory installed-package-recipes.txt has name of recipe, but incorrect pointer to what is actual full version of recipe or full version of binary package.
That is important e.g. for npm packages.

kad:
It would be better to actually enhance oe-pkgdata-util to produce actual line with info instead of multiple call, but for now it's ok even that way, until better functionality available.

pohly:
Looks okay to me now. I assume the change in the installed-package-recipes.txt content will be dealt with in the tools parsing the data?

I'd prefer to have more information in the commit message about what's changing in the file (say, with a specific example) and why that needs to be fixed, because it is not obvious, but that's not a blocker for merging.

kad:
It will be anyway re-adjusted if/when http://lists.openembedded.org/pipermail/openembedded-core/2016-June/122487.html will be merged

PR #198, opened 2016-06-06
iotivity-node: remove obsolete signed/unsigned char patch

nagineni:
Build failure has already been fixed by updating the iotivity-node
to latest, so this patch is not needed anymore.

Signed-off-by: Sudarsana Nagineni <sudarsana.nagineni@intel.com>


nagineni:
@pohly

pohly:
Looks good. I was wondering briefly why we didn't get build errors because of the patch, until I saw that I had added it conditionally based on the revision ;-} So the patch has been unused already for a while.

Let's merge after a clean test build (just to be sure).

nagineni:
> Looks good. I was wondering briefly why we didn't get build errors because of the patch, until I saw that I had added it conditionally based on the revision ;-} So the patch has been unused already for a while.

You are right. :)

> Let's merge after a clean test build (just to be sure).

Build bots are green. Can you please merge this?

PR #199, opened 2016-06-06
Clean up .vdi image options that are now .ova image options

dbkinder:
We're no longer creating .vdi images (for VirtualBox) in favor of
using Open Virtualization Archive (.ova) files,  a pre-packaged appliance file
that can be directly imported to VirtualBox and other virtualization applications.

We updated the .vdi to .ova change in the documentation (fixed in PR#194)
and this PR removes the remaining vestiges in configuration files.


meta-ostro/classes/ostro-image.bbclass
```diff
@@ -230,15 +230,15 @@ ostro_root_authorized_keys () {
 NOISO = "1"
 
 # Add support for compressing images with zip. Works for arbitrary image
-# types. Example: OSTRO_VM_IMAGE_TYPES = "dsk.zip dsk.vdi.zip"
```
gvancuts: Does that get rid of the possibility to create a VDI file directly (should one wish to)?

kad: we probably don't want to. if there is ova, there is no need on vdi.
Or, if somebody don't want to use ova but still use VBox, he should have enough knowledge to how to convert between raw images to something else.

gvancuts: OK

PR #200, opened 2016-06-08
classes: Python3 compatibility

pohly:
The old "except Foo, ex" no longer works once bitbake migrates to
Python3.

This is required (but not sufficient) for solving the Python3-related
problems in https://github.com/ostroproject/ostro-os/pull/116

We also need some of the already pending meta-openembedded patches.


kad:
👍

pohly:
test this please

PR #201, opened 2016-06-08
Documentation for deploying to internal storage.

igor-stoppa:
This section shows how to deploy an Ostro OS image
to the internal storage that might be available on certain
platforms.

Signed-off-by: Igor Stoppa <igor.stoppa@intel.com>

### DIFF
--- a/doc/howtos/booting-and-installation.rst
+++ b/doc/howtos/booting-and-installation.rst
@@ -127,6 +127,62 @@ includes a checksum verification, you can also use the traditional :command:`dd`
 Unplug the removable media from your development system and you're ready to plug
 it into your target system.
 
+Installing to internal media on generic x86 EFI platforms
+=========================================================
+
+This section applies to those cases where the device is x86 compatible (ex: Gigabyte NUC).
+The simpler way is to follow these steps:
+
+#. Create a Bootable Media using both an image that is compatible with the device
+   and a supported removable media (SD card or USB stick).
+#. Boot the device with the removable media created at the previous step.
+#. Once the device is booted, identify the booted media, for example by executing::
+
+      # mount
+
+   and locating in its output which partition is mounted ``on /``.
+   This will be a partition belonging to the source block device.
+   Example::
+
+      /dev/sdb3 on / type ext4 (rw,realtime,i_version,data=ordered)
+
+   Shows that the root partition is on ``/dev/sdb3``, therefore the source block device will
+   be ``/dev/sdb``.
+#. Identify the block device representing the internal storage.
+   This part is really specific to each platfrom: typically the internal storage is
+   associated to ``/dev/sda`` for hard disks and SSD units and to ``/dev/mmcblk0`` for eMMC
+   devices, but it should be confirmed against the board manual.
+#. Issue the ``dd`` command with the parameters identified in the last 2 points.
+   Example with internal storage as eMMC on ``/dev/mmcblk0`` and removable media as USB stick
+   on ``/dev/sda``::
+
+      # dd if=/dev/sda of=/dev/mmcblk0 bs=5M && sync
+
+#. When the command has terminated, poweroff the device, extract the removable media
+   and power it back on. This should be sufficient to have the unit now using the internal
+   storage.
+
+
+Removing Ostro OS from internal media
+=====================================
+   Ostro OS uses GID for identifying the rootfs and the GID is kept consistent across all the
+   images produced. This means that trying to boot a system with 2 Ostro OS images available at the
+   same time will likely produce unwanted/undetermined results.
+   Before trying again to boot from a removable media, the internal media should be wiped.
+   To achieve this:
+#. Boot from the internal media.
+#. Identify the root block device containing the rootfs partition.
+#. Wipe a sufficient part of that block device to make it unbootable.
+   In practice it means wiping the primary GPT and at least the beginngin of both partitions that
+   might be EFI-bootable. In an unmodified Ostro disk layout, this means the first 2 partitions,
+   which are 15MB each. 20 MB would be sufficient to cover the beginning of the disk, the first
+   partition and the beginning of the second. 30MB will wipe even the beginning of the rootfs.
+   Example with rootfs on ``/dev/sda3``::
+
+      # dd if=/dev/zero of=/dev/sda bs=5M count=6 && sync
+
+#. Poweroff the device (ignore possible ext4 error messages).
+#. Insert the removable media with the new Ostro OS image and power on.
 
 MinnowBoard Turbot - a MinnowBoard MAX Compatible
 =================================================
--- a/meta-ostro/classes/buildhistory-extra.bbclass
+++ b/meta-ostro/classes/buildhistory-extra.bbclass
@@ -120,21 +120,33 @@ python buildhistory_extra_emit_kernelconfig() {
 }
 
 buildhistory_get_image_installed_append() {
-	# Create a file mapping installed packages to recipes
-	printf "" > ${BUILDHISTORY_DIR_IMAGE}/installed-package-recipes.txt
-	cat ${IMAGE_MANIFEST} | while read pkg pkgarch version
-	do
-		if [ -n "$pkg" ] ; then
-			recipe=`oe-pkgdata-util -p ${PKGDATA_DIR} lookup-recipe $pkg`
-			pkge=`oe-pkgdata-util -p ${PKGDATA_DIR} read-value PKGE $pkg`
-			if [ "$pkge" != "" ] ; then
-				pkge="$pkge-"
-			fi
-			pkgr=`oe-pkgdata-util -p ${PKGDATA_DIR} read-value PKGR $pkg`
-			if [ "$pkgr" != "" ] ; then
-				pkgr="-$pkgr"
-			fi
-			echo "$pkg $version $pkge$version$pkgr $recipe" >> ${BUILDHISTORY_DIR_IMAGE}/installed-package-recipes.txt
-		fi
-	done
+    # Create a file mapping installed packages to recipes
+    printf "" > ${BUILDHISTORY_DIR_IMAGE}/installed-package-recipes.txt
+    export PSEUDO_UNLOAD=1
+    cat ${IMAGE_MANIFEST} | while read pkg pkgarch version
+    do
+        if [ -n "$pkg" ] ; then
+            pkgv=`oe-pkgdata-util -p ${PKGDATA_DIR} read-value PKGV $pkg`
+            pkge=`oe-pkgdata-util -p ${PKGDATA_DIR} read-value PKGE $pkg`
+            if [ "${pkge}" != "" ] ; then
+                pkge="${pkge}:"
+            fi
+            pkgr=`oe-pkgdata-util -p ${PKGDATA_DIR} read-value PKGR $pkg`
+            if [ "${pkgr}" != "" ] ; then
+                pkgr="-${pkgr}"
+            fi
+            recipe=`oe-pkgdata-util -p ${PKGDATA_DIR} read-value PN $pkg`
+            recipev=`oe-pkgdata-util -p ${PKGDATA_DIR} read-value PV $pkg`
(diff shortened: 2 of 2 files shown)


igor-stoppa:
@dbkinder @gvancuts @uhofemei @kad @pohly please review

kad:
maybe it makes sense to sync + remount / as read-only during dd operation ? it's not the best solution, but a bit better than possibility that something gets changed during copy operation ?

igor-stoppa:
It doesn't really matter, because that is no different from what might
happen if there is a power loss during disk write. Ext4 is supposed to deal
with it.

I preferred to omit it, in favour of keeping the command short.

On 8 June 2016 at 14:42, Alexander D. Kanevskiy <notifications@github.com>
wrote:

> maybe it makes sense to sync + remount / as read-only during dd operation
> ? it's not the best solution, but a bit better than possibility that
> something gets changed during copy operation ?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/ostroproject/meta-ostro/pull/201#issuecomment-224564865>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AA2GSaysbKVxaCgsCOgUlDtGlcVodYa-ks5qJqqZgaJpZM4Iw3ZM>
> .
>



-- 
igor

kad:
ok for me then.

doc/howtos/booting-and-installation.rst
```diff
@@ -127,6 +127,58 @@ includes a checksum verification, you can also use the traditional :command:`dd`
 Unplug the removable media from your development system and you're ready to plug
 it into your target system.
 
+Installing to internal media on generic x86 EFI platforms
+=========================================================
+
+This section applies to those cases where the device is x86 compatible (ex: Gigabyte NUC).
+The simpler way is to follow these steps:
+
+#. Create a Bootable Media using an both an image that is compatible with the device
```
gvancuts: Can you remove the extra 'an' (...using an both an image...)?

doc/howtos/booting-and-installation.rst
```diff
@@ -127,6 +127,58 @@ includes a checksum verification, you can also use the traditional :command:`dd`
...
+   and a supported removable media (SD card or USB stick).
+#. Boot the device with the removable media created at the previous step.
+#. Once the device is booted, identify the booted media, for example by executing::
+
+      $ mount
```
gvancuts: In a dev image, we're logged in as root so the the prompt is '#' (ok, I know I'm splitting hairs here...)

doc/howtos/booting-and-installation.rst
```diff
@@ -127,6 +127,58 @@ includes a checksum verification, you can also use the traditional :command:`dd`
...
+
+   and locating in its output which partition is mounted ``on /``.
+   This will be a partition belonging to the source block device.
+   Example::
+
+      /dev/sdb3 on / type ext4 (rw,realtime,i_version,data=ordered)
+
+   Shows that the root partition is on `/dev/sdb3`, therefore the source block device will
+   be `/dev/sdb`.
```
gvancuts: I think you need double-backticks here.

doc/howtos/booting-and-installation.rst
```diff
@@ -127,6 +127,58 @@ includes a checksum verification, you can also use the traditional :command:`dd`
...
+#. Identify the block device representing the internal storage.
+   This part is really specific to each platfrom: typically the internal storage is
+   associated to `/dev/sda` for hard disks and SSD units and to `/dev/mmcblk0` for eMMC
+   devices, but it should be confirmed against the board manual.
```
gvancuts: Would such info really be found in a board manual? Or should we instead recommend to run a command such as blkid or similar?

doc/howtos/booting-and-installation.rst
```diff
@@ -127,6 +127,58 @@ includes a checksum verification, you can also use the traditional :command:`dd`
...
+#. Issue the `dd` command with the parameters identified in the last 2 points.
+   Example with internal storage as eMMC on `/dev/mmcblk0` and removable media as USB stick
+   on `/dev/sda`::
```
gvancuts: Double-backticks

doc/howtos/booting-and-installation.rst
```diff
@@ -127,6 +127,58 @@ includes a checksum verification, you can also use the traditional :command:`dd`
...
+
+      $ dd if=/dev/sda of=/dev/mmcblk0 bs=5M && sync
```
gvancuts: Same nitpick as above, the prompt will be '#'

doc/howtos/booting-and-installation.rst
```diff
@@ -127,6 +127,58 @@ includes a checksum verification, you can also use the traditional :command:`dd`
...
+
+#. When the command has terminated, poweroff the device, extract the removable media
+   and power it back on. This should be sufficient to have the unit now using the internal
+   storage.
+
+
+Removing Ostro OS from internal media
+=====================================
+   Ostro OS uses GID for identifying the rootfs and the GID is kept consistent across all the
+   images produced. This means that trying to boot a system with 2 Ostro OC images avilable at the
```
gvancuts: s/OC/OS?

gvancuts: s/avilable/available

doc/howtos/booting-and-installation.rst
```diff
@@ -127,6 +127,58 @@ includes a checksum verification, you can also use the traditional :command:`dd`
...
+   same time will likely produce unwanted/undetermined results.
+   Before trying again to boot from a removable media, the internal media should be wiped.
+   To achieve this:
+#. Boot from the internal media.
+#. Identify the root block device containing the rootfs partition.
+#. Wipe a sufficient part of that block device.
+   Example with rootfs on `/dev/sda3`::
```
gvancuts: Double-backticks.

doc/howtos/booting-and-installation.rst
```diff
@@ -127,6 +127,58 @@ includes a checksum verification, you can also use the traditional :command:`dd`
...
+Removing Ostro OS from internal media
+=====================================
+   Ostro OS uses GID for identifying the rootfs and the GID is kept consistent across all the
+   images produced. This means that trying to boot a system with 2 Ostro OC images avilable at the
+   same time will likely produce unwanted/undetermined results.
+   Before trying again to boot from a removable media, the internal media should be wiped.
+   To achieve this:
+#. Boot from the internal media.
+#. Identify the root block device containing the rootfs partition.
+#. Wipe a sufficient part of that block device.
```
gvancuts: Can you perhaps expand on what 'sufficient' means? In the command below, I think what you're doing is essentially wipe the GPT and that's it, is it correct?

doc/howtos/booting-and-installation.rst
igor-stoppa: done

doc/howtos/booting-and-installation.rst
igor-stoppa: done

doc/howtos/booting-and-installation.rst
igor-stoppa: done

doc/howtos/booting-and-installation.rst
igor-stoppa: I don't know. Maybe lspci & lsusb would help. But not blkid, afaik it relies on logical information present on the block devices. But they might be empty/corrupted. In reality I expect one knows what local storage is present on a certain device.

doc/howtos/booting-and-installation.rst
igor-stoppa: done

doc/howtos/booting-and-installation.rst
igor-stoppa: done

doc/howtos/booting-and-installation.rst
igor-stoppa: done, both

doc/howtos/booting-and-installation.rst
igor-stoppa: No, that would not be sufficient, because there is a backup of the GPT at the "end" of the disk.
In practice this backup would work only if theh backup GPT is really at theend of the phyisical disk, but there would be some cases where it would still manage to start. What it really does here, is to make the disk unbootable by wiping the beginning of the EFI partition. But there is a problem, I forgot that the bootable partition might be the 2nd, after a swupd, so I'll fix that part.

doc/howtos/booting-and-installation.rst
igor-stoppa: done

igor-stoppa:
I couldn't get rid of a spurious patch that was creeping in, no matter what I did, so I removed the branch.

igor-stoppa:
The quest continues here: https://github.com/ostroproject/meta-ostro/pull/203

pohly:
@igor-stoppa "git rebase -i origin/master" and deleting the unwanted commit should have worked. Perhaps try that the next time.

igor-stoppa:
Eventually with @juimonen we figured it out: I was doing an interactive rebase to squash the modification to the PR, based on the comments received, but I was including in the history also a merge (without touching it, though), but this was sufficient to turn that merge entry into the real patch. I suppose it's a feature of git. Lesson learned: don't get even close to merges.

pohly:
@igor-stoppa so you weren't rebasing onto origin/master? Typically rebase flattens the history, so I'd expect it to get rid of any extra merge commits. Just trying to broaden my understanding of what works and what doesn't with git here ;-}

igor-stoppa:
@pohly why would I have to rebase?
I clone the meta-ostro, create a branch for my patch, write the patch, push it to my repo, thne create the PR
After I get feedback, I create a new patch and squash it onto the one I wrote earlier, then force push to my forked repo.

Where would I rebase?

gvancuts:
You should rebase if a commit was made to meta-ostro in between the first and second rev of your patch. At least, that's what we recommend here: https://ostroproject.org/documentation/quick_start/contributor-guide.html#contributor-guide :-)

igor-stoppa:
But it wasn't!


What was happening is the following:

$ git log:

* my latest patch to be squashed
* older version of the patch
* merge commit


then i did:

$ git rebase -i HEAD~3

pick xxxx merge commit
pick yyyy old patch
pick zzzz new patch

and I did

pick xxxx merge commit
pick yyyy old patch
f zzzz new patch

and this turned into:

$git log

* squashed new and old patch
* patch that was merged, instead of the merge commit  <---!!!!!!!

So what seems to work is to *not* do 

git rebase -i HEAD~arbitrary number

but instead 

git rebase -i HEAD~2

igor-stoppa:
@pohly btw, the branch in use was created by doing
git checkout master -B doc

right after cloning -was I supposed to do a rebase here?

but even if I did, the master branch in github still contains the merge commits, so I would end up in the same situation, wouldn't I? That I have in my branch the flattened history, while upstream has merge commits.

pohly:
@igor-stoppa I'm no longer able to follow what you did. Perhaps your "master" branch was not the same as "origin/master". Also, in your example, "git rebase -i HEAD~3" is not the same as "git rebase -i origin/master": the former rebases onto the parent of the merge commit and thus preserves that parent commit (i.e. doesn't allow you to remove it), the latter onto to the target of your PR (i.e. rebases all commits that are included in your PR, and allows you to remove the ones you don't want). I still think the latter would have worked.

But there's no point in discussing this further - next time when there is an issue, we can investigate based on the actual commits instead of speculating.

pohly:
Actually, it just occurred to me what you probably did: with "git rebase -i HEAD~3" you ended up rebasing some of the origin/master branch, leading to re-committing already merged commits which then ended up in your PR. That's why GitHub now shows 73b71b23 as "kad committed with igor-stoppa 3 days ago" - kad is the author and you the committer.

"git rebase origin/master" avoids that because one does not have to determine manually what the right number after HEAD~ is.

igor-stoppa:
@pohly No, Ithink I now understood what you meant with "git rebase -i origin/master" and will try it the next time. It's just that I was not expecting that picking the merge commit in my HEAD~3 would actually alter it and flatten it.

PR #202, opened 2016-06-08
ostro os: remove questionable zip and iotop

pohly:
Internal review raised some concerns about the upstream maintenance of
these components. Because there are no hard requirements for them and
there are alternatives for interactive use (tar is better for archival
than zip, and atop can show similar information as iotop), the easiest
solution is to remove the components.


kad:
will it prevent of building zip-native which is needed for image conversions ?

pohly:
"zip-native" will still be available, without having to be listed. supported-recipes.bbclass only deals with components compiled for the target, because that is our main focus regarding the component selection.

pohly:
@kad okay to merge? If so, please hit "merge pull request" or document that I may merge it myself.

kad:
ok to merge.

PR #204, opened 2016-06-08
local.conf.sample: add option to enable systemd-bootchart

gvancuts:
- Add full path to a building-images.rst reference in local.conf
- Add paragraph that provides an option to enable systemd-bootchart
and add some basic explanation on how to use it

[skip ci]

### DIFF
--- a/meta-ostro/conf/local.conf.sample
+++ b/meta-ostro/conf/local.conf.sample
@@ -266,7 +266,7 @@ CONF_VERSION = "3"
 # By default, only recipes explicitly declared to be okay for building Ostro OS
 # are allowed to be built. Adding or depending on anything else triggers a
 # message saying "The following unsupported recipes are required for the build"
-# and aborts. See building-images.rst for more information.
+# and aborts. See doc/howtows/building-images.rst for more information.
 SUPPORTED_RECIPES_CHECK = "fatal"
 
 # Uncomment the following line to disable the check:
@@ -290,6 +290,21 @@ SUPPORTED_RECIPES_CHECK = "fatal"
 # and uncomment the following line.
 #
 #require conf/distro/include/ostro-os-production.inc
+
+# systemd-bootchart is a useful tool to analyze and optimize a system
+# boot time. The tool is available in Ostro and needs to be activated
+# by a kernel command-line parameter which requires to build a new
+# image. This tool also requires some kernel configurations to be active,
+# specifically CONFIG_DEBUG_KERNEL and CONFIG_SCHEDSTATS.
+#
+# To use systemd-bootchart with the intel-corei7-64 or intel-quark machine
+# uncomment the two lines below. Other MACHINEs that use u-boot (such as
+# edison and beaglebone) use a different mechanism to modify the kernel
+# command-line parameter, this is not covered here. The bootchart graphs
+# will be available in /run/log/.
+#
+#APPEND_append = " init=/lib/systemd/systemd-bootchart"
+#SRC_URI_append_pn-linux-yocto = " file://bootchart.cfg"
 
 #
 # Misc configuration


kad:
👍

meta-ostro/conf/local.conf.sample
```diff
@@ -291,6 +291,18 @@ SUPPORTED_RECIPES_CHECK = "fatal"
 #
 #require conf/distro/include/ostro-os-production.inc
 
+# systemd-bootchart is a useful tool to analyze and optimize a system
+# boot time. The tool is available in Ostro and needs to be activated
+# by a kernel command-line parameter which requires to build a new
+# image. You also need to make sure that you enable CONFIG_SCHEDSTATS=y
+# (that will also set CONFIG_SCHED_INFO=y) in your kernel, this option
+# is not enabled by default. For instructions on how to modify an Ostro
+# kernel, please check doc/howtos/modifying-ostro-kernel.rst.
```
pohly: We could make it simpler to add CONFIG_SCHEDSTATS=y by providing a kernel config fragment for it. See meta-intel-iot-security/meta-security-framework/recipes-kernel/linux/linux-%.bbappend for an example where that is done for enabling audit.

Just a thought. We can also merge this as it is and enhance it later - @gvancuts, your call as the author of the PR.

gvancuts: I very much like this idea, I did not know how to do something like this via local.conf. As I feel there is no urgency to merge this one, I'd propose to keep this open and I'll figure out how to make this simpler for customers. Would it be OK to do it as follows?
- Add fragment to meta-ostro-bsp/recipes-kernel/linux-yocto/linux-yocto
- Add #SRC_URI_append = " file://bootchart.cfg" to local.conf.sample

gvancuts: Scrap that, I need to figure out the mechanism better, just let me know where the fragment should go?

pohly: meta-ostro-bsp/recipes-kernel/linux/ might be the right place, similar to meta-intel-iot-security/meta-security-framework/recipes-kernel/linux. Just remember that it cannot assume that the kernel is called "linux-yocto". We also have "linux-yocto-edison", so a "linux-yocto_*.bbappend" does not work for all kernel, and some downstream user might use something which is not a Yocto kernel at all. "linux-*.bbappend" as in meta-intel-iot-security works, but is a bit hacky (see commit log messages there).

The alternative to the linux-*.bbappend would be to extend the file search path in the layer conf for all recipes, but that might have unexpected consequences, so I'd not go down that route.

gvancuts: Here is what I have tried so far:
- Add a fragment file to my ostro-os repo
   gvancuts@iotdev:~/ostro-os$ cat meta-ostro-bsp/recipes-kernel/linux-yocto/linux-yocto/bootchart.cfg
   CONFIG_SCHEDSTATS=y
- Modify my local.conf:
   # Add a command-line parameter
   APPEND_append = " init=/lib/systemd/systemd-bootchart"
   SRC_URI_append_pn-linux-yocto = " file://bootchart.cfg"

I have tried this for intel-quark so far (testing images in VirtualBox), the weird thing is that the kernel does have SCHEDSTAT enabled and the cmdline (as checked with 'less /proc/cmdline' on a running system has the extra init command - see below). However, there is no svg file created in /run/log

If I manually start /lib/systemd/systemd-bootchart, then such file is created as expected.

Any idea what is going wrong? Does the order of kernel command-line parameters matter?

root@intel-quark:~# cat /proc/cmdline
 console=ttyS1,115200 console=tty0 init=/lib/systemd/systemd-bootchart fsck.mode=skip rootflags=i_version ro root=PARTUUID=12345678-9abc-def0-0fed-cba987654321 rootfstype=ext4

About the fact we have different kernel, I was thinking we could have multiple lines, one for each kernel (PN) in use in Ostro and with a simple note that should someone use yet a different kernel, they should adjust that very line. Does that sound like a good approach?

uhofemei: The man page says systemd-bootchart is in /usr/lib, so init=/usr/lib/systemd/systemd-bootchart

Ulf

gvancuts: It's installed in /lib in Ostro, not /usr/lib

barbieri: as far as I remember, the boot cmdline is just to collect the stuff. You
still need to manually ask for the svg to be rendered, but I may be too old
fashioned :-)

On Fri, Jun 10, 2016 at 11:18 AM, Geoffroy Van Cutsem <
notifications@github.com> wrote:

> In meta-ostro/conf/local.conf.sample
> <https://github.com/ostroproject/meta-ostro/pull/204#discussion_r66620359>
> :
>
> > @@ -291,6 +291,18 @@ SUPPORTED_RECIPES_CHECK = "fatal"
> >  #
> >  #require conf/distro/include/ostro-os-production.inc
> >
> > +# systemd-bootchart is a useful tool to analyze and optimize a system
> > +# boot time. The tool is available in Ostro and needs to be activated
> > +# by a kernel command-line parameter which requires to build a new
> > +# image. You also need to make sure that you enable CONFIG_SCHEDSTATS=y
> > +# (that will also set CONFIG_SCHED_INFO=y) in your kernel, this option
> > +# is not enabled by default. For instructions on how to modify an Ostro
> > +# kernel, please check doc/howtos/modifying-ostro-kernel.rst.
>
> It's installed in /lib in Ostro, not /usr/lib
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/ostroproject/meta-ostro/pull/204/files/918986fe9b4e4ec9d104c5fcd93e8c0e5db7172d#r66620359>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAAHLqJeh6FgAb6b78-63lkLzAM5UqQrks5qKXIcgaJpZM4Iw_6M>
> .
>



-- 
Gustavo Sverzut Barbieri
Mobile: +55 (19) 99225-2202

gvancuts: I wasn't trying to render it, the problem is that it's not even created. It does get created if I run /lib/systemd/systemd-bootchart manually though... weird. I was wondering whether there could be some timing issue when the various parts of the filesystem get mounted but I'll take any other idea/recommendation.

gvancuts: OK, it looks like I just wasn't patient enough... ;-/
It does get created but it takes a little while to complete (20 sec with the default settings).

I have verified this on both intel-quark and intel-corei7-64 so far, I will update this PR to includes the lines (commented out by default) that include the fragment file. The PR with the fragments will be sent in a separate PR.

gvancuts:
Here is the latest status on my investigation using the following technique:
* Add a fragment file to my ostro-os repo
   gvancuts@iotdev:~/ostro-os$ cat meta-ostro-bsp/recipes-kernel/linux-yocto/linux-yocto/bootchart.cfg
   CONFIG_DEBUG_KERNEL=y
   CONFIG_SCHEDSTATS=y

* Modify my local.conf:
   # Add a command-line parameter and kernel config fragment
   APPEND_append = " init=/lib/systemd/systemd-bootchart"
   SRC_URI_append_pn-linux-yocto = " file://bootchart.cfg"

This works well for intel-quark and intel-corei7-64. It does not, however work for intel-edison because it uses u-boot and the APPEND variable does not have any effect. I haven't tried on BeagleBone but I suspect it won't work either for the same reason (APPEND does not apply to u-boot?).

I'm thinking I will submit a new patch (+ another one with the config fragment and indicate in the note that it only works for intel-galileo and intel-corei7-64 machines.

gvancuts:
A new patch has been force-pushed. It goes hand-in-hand with https://github.com/ostroproject/meta-ostro-bsp/pull/50.

gvancuts:
Hi @pohly , am I supposed to remove the 'author-rework-required' label myself (if so, how do I do that)?

pohly:
@gvancuts unfortunately you can't remove that label, which makes the labels a lot less useful than originally hoped. I'll remove the label. I was waiting for @jlaako to merge https://github.com/ostroproject/meta-ostro-bsp/pull/50

PR #205, opened 2016-06-08
supported recipes: extension mechanism for report generator

pohly:
Besides the new functionality, a .gitgnore file gets added (useful
when editing and using the layer directly, as I did when working on the
feature), and Python3 compatibility gets addressed.

The initial commit is also part of PR #200, which should be merged first.

This currently fails to pass combo-layer because the .gitignore file overlaps on the combined repo with the ostro-os/.gitignore. I will update combo-layer.conf to fix that.


meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
+# Python code implementing most of the logic behind
+# supported-recipes.bbclass.
+
+import bb
```
kad: should be imported after system modules

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+import csv
+try:
+    from exceptions import RuntimeError
```
kad: RuntimeError present in global scope in both py2 and py3.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+except ImportError:
+    import urllib.parse as urlparse
+import pkgutil
+import inspect
+
+try:
+    import report
+except ImportError:
+    import supportedrecipes.report as report
+
```
kad: PEP 328, relative imports ?

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+class Columns:
```
kad: class Columns(object)

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+    '''Base class for all classes which add additional columns to the SUPPORTED_RECIPES_SOURCES report.'''
+    def __init__(d, all_rows):
```
kad: PEP8, docstrings.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+        pass
+
+    def extend_header(row_headers):
+        '''Called with a list of field names, in the order in which the
+        resultig .cvs report will have them.  extend_header() then may
+        extend the list of fields. See supportedrecipes/__init__.py for
+        a list of already present fields.
+        '''
+        pass
```
kad: PEP8, docstrings.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def extend_row(row):
+        '''Called with a hash containing one line entry. Can add more fields.'''
+        pass
```
kad: PEP8.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+
+def load_supported_recipes(d):
+    class SupportedRecipe:
```
kad: (object).

kad: no need to nest this class definition inside function. Unmaintainable code.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+            self.pattern = pattern
+            self.linenumber = linenumber
+            parts = pattern.split('@')
+            if len(parts) != 2:
+                raise RuntimeError("%s.%d: entry must have format <recipe name regex>@<collection name regex>, splitting by @ found %d parts instead: %s" % (
+                    filename,
+                    linenumber,
+                    len(parts),
+                    pattern))
+            def parse_regex(regex):
```
kad: no need to be inline function. move out of 3 nested scopes.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+                        str(ex)))
+            self.pn_re = parse_regex(parts[0])
+            self.collection_re = parse_regex(parts[1])
+
+        def supported(self, pn, collection):
+            supported = bool((pn is None or self.pn_re[0].match(pn)) and
+                (collection is None or self.collection_re[0].match(collection)))
+            return supported
+
+    class SupportedRecipes:
```
kad: (object).
No need to be scoped inside function.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+                    return True
+            return False
+
+        def find_collections(self, pn):
+            collections = set()
+            for supported_recipe in self.supported:
+                if supported_recipe.supported(pn, None):
+                    collections.add(supported_recipe.collection_re[1])
+            return collections
+
```
kad: can be onliner, where find_collections called (only in one place):
```python
 {supported_recipe.collection_re[1] for supported_recipe in self.supported if supported_recipe.supported(pn, None)}
```

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+    files = []
+    supported_files = d.getVar('SUPPORTED_RECIPES', True)
+    if not supported_files:
+        bb.fatal('SUPPORTED_RECIPES is not set')
+    supported_recipes = SupportedRecipes()
+    for filename in supported_files.split():
+        try:
+            with open(filename) as f:
+                linenumber = 1
+                for line in f.readlines():
```
kad:
```
for line in f:
```
readlines() would read whole file in memory instead of iterating over it.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+    sources = []
+    for url in src:
+        scheme, netloc, path, query, fragment = urlparse.urlsplit(url)
+        if scheme != 'file':
+            parts = path.split(';')
+            if len(parts) > 1:
+                path = parts[0]
+                params = dict([x.split('=') if '=' in x else (x, '') for x in parts[1:]])
+            else:
+                params = {}
```
kad: manual parsing of params instead of urlparse() ?

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+            sources.append((name, '%s://%s%s' % (scheme, netloc, path)))
+    dumpfile = d.getVar('SUPPORTED_RECIPES_SOURCES_DIR', True) + '/' + pn + filename
+    bb.utils.mkdirhier(os.path.dirname(dumpfile))
+    with open(dumpfile, 'w') as f:
+        # File intentionally kept small by not writing a header line. Guaranteed to contain SOURCE_FIELDS.
+        writer = csv.writer(f)
+        for idx, val in enumerate(sources):
+            name, url = val
+            if name and not len(sources) == 1:
+                fullname = '%s/%s' % (pn, name)
```
kad: os.path.join ?

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+    report_sources = d.getVar('SUPPORTED_RECIPES_SOURCES', True)
+
+    unsupported = {}
+    sources = []
+    for pn, pndata in depgraph['pn'].iteritems():
+        # We only care about recipes compiled for the target.
+        # Most native ones can be detected reliably because they inherit native.bbclass,
+        # but some special cases have to be hard-coded.
+        # Image recipes also do not matter.
+        def isnative():
```
kad: redefining function on every loop run, doesn't make sense. move to outer scope that function.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+            if isnative_exception.match(pn):
+                return True
+        if not isnative():
+            filename = pndata['filename']
+            collection = bb.utils.get_file_layer(filename, d)
+            supported = supported_recipes.recipe_supported(pn, collection)
+            if not supported:
+                unsupported[pn] = collection
+            if report_sources:
+                dumpfile = dirname + '/' + pn + filename
```
kad: os.path.join

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+                with open(dumpfile) as f:
+                    reader = csv.reader(f)
+                    for row in reader:
+                        row_hash = dict([(f, row[i]) for i, f in enumerate(SOURCE_FIELDS)])
```
kad:
```python
row_hash = { f: row[i] for i, f in enumerate(SOURCE_FIELDS) }
```
instead of list to dict conversion.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+                        row_hash['supported'] = 'yes' if supported else 'no'
+                        sources.append(row_hash)
+
+    if report_sources:
+        with open(report_sources, 'w') as f:
+            fields = SOURCE_FIELDS
+            # Insert after 'collection'.
+            fields.insert(fields.index('collection') + 1, 'supported')
```
kad: this would modify global constant instead of local copy "fields".
```python
In [1]: SOURCE_FIELDS="a,b,c,d".split(",")
In [2]: f=SOURCE_FIELDS
In [3]: f.insert(f.index("c")+1, "something")
In [4]: SOURCE_FIELDS
Out[4]: ['a', 'b', 'c', 'something', 'd']
```

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+        deps = {}
+        for task, taskdeps in depgraph['tdepends'].iteritems():
+            pn = task.split('.')[0]
+            pndeps = deps.setdefault(pn, set())
+            for taskdep in taskdeps:
+                pndep = taskdep.split('.')[0]
+                if pndep != pn:
+                    pndeps.add(pndep)
+                    roots.discard(pndep)
+        for pn in deps.keys():
```
kad: it's enough "for pn in deps"

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+            deps[pn] = sorted(deps[pn])
+
+        # We can prune the search tree a lot by keeping track of those recipes which are already
+        # known to not depend on an unsupported recipe.
+        okay = set()
+
+        class TruncatedError(Exception):
```
kad: defining exception within function?

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+
+        truncated = False
+        try:
+            for pn in sorted(roots):
+                visit_recipe(pn)
+        except TruncatedError:
+            truncated = True
+
+        def collection_hint(pn):
+            '''Determines whether the recipe would be supported in some other collection.'''
```
kad: PEP8

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+            collections = supported_recipes.find_collections(pn)
+            return ' (would be supported in %s)' % ' '.join(collections) if collections else ''
```
kad: could be moved to outer scope, to not clutter inner function scope.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+
+        logger('The following unsupported recipes are required for the build:\n  ',
+               # In contrast to re.escape(), we only escape + (as in gtk+3).
+               # Escaping all non-alphanumerics makes many entries (like linux-yocto)
+               # unnecessarily less readable (linux\-yocto).
+               '\n  '.join(sorted(['%s@%s%s' % (pn.replace('+', r'\+'), collection.replace('+', r'\+'), collection_hint(pn)) for pn, collection in unsupported.iteritems()])),
```
kad: rewrite that line to be more maintainable.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
```diff
@@ -0,0 +1,346 @@
...
+Here are the dependency chains (including DEPENDS and RDEPENDS)
+which include one or more of the unsupported recipes. -> means "depends on"
+and * marks unsupported recipes:
+  ''' if lines else '',
+               '\n  '.join([' -> '.join([('*' if pn in unsupported else '') + pn for pn in line]) for line in lines]),
+               ('''
+  ...
+  Output truncated, to see more increase SUPPORTED_RECIPES_CHECK_DEPENDENCY_LINES
+  (currently %d).''' % max_lines) if truncated else '',
+               '''
```
kad: Unreadable piece of code :(

meta-ostro/lib/supportedrecipes/supportedrecipes.py
pohly: Interesting. If I remember correctly, pylint complained about "unknown class" when taken from global scope. I'll check.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
pohly: Agreed. That's a leftover from the time when this was in a .bbclass.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
pohly: Not in this case, pn and name are not filename components. This implements the <recipe>[/<source name>] special case for recipes with multiple components.

meta-ostro/lib/supportedrecipes/supportedrecipes.py
pohly: That was intentionally, because the outcome of urlparse() is a bit unpredictable (sometimes it splits off params, sometimes it doesn't). We have a workaround for that (re-do the splitting after adding the scheme to the list of schemes for which splitting is expected), but I'm uncertain whether modifying the (global?) urlparse behavior in this context would be the right thing to do.

I can change it, if you want.

pohly:
@kad added cleanups, please re-review.

pohly:
I also added fixes for Python3 compatibility here, to avoid conflicts that I would get when making the same change independently in the supported-recipes.bbclass on meta-ostro master.

However, this means that this PR is needed before https://github.com/ostroproject/ostro-os/pull/116 can proceed.

meta-ostro/lib/supportedrecipes.py
```diff
@@ -0,0 +1,388 @@
...
+    sources = []
+    for url in src:
+        scheme, netloc, path, query, fragment = urlparse.urlsplit(url)
+        if scheme != 'file':
+            parts = path.split(';')
+            if len(parts) > 1:
+                path = parts[0]
+                params = dict([x.split('=') if '=' in x else (x, '') for x in parts[1:]])
+            else:
+                params = {}
```
kad: using urlparse params split would be still better, as even in global scope, programs can't be dependant on list of schemas supported by default in the lib (it differs between python versions anyway). Not critical at the moment, but would be better.

kad:
looks good, except small non-critical thing with urlparse and params.

pohly:
retest this please

PR #206, opened 2016-06-09
prepare for Python3 migration

pohly:
This fixes build issues and allows Python3 in the target images.

Blocks https://github.com/ostroproject/ostro-os/pull/116 but is safe to merge already now.

### DIFF
--- a/meta-ostro/conf/distro/ostro.conf
+++ b/meta-ostro/conf/distro/ostro.conf
@@ -68,7 +68,7 @@ SIGGEN_LOCKEDSIGS_TASKSIG_CHECK = "none"
 # For more predictable results is possible to set OS_RELEASE manually,
 # either in local.conf or in the environment like this:
 #   BB_ENV_EXTRAWHITE="$BB_ENV_EXTRAWHITE OS_VERSION" OS_VERSION=110 bitbake ...
-OS_VERSION ?= "${@ str(((int('${DATETIME}') - 20160000000000) / 100) * 10) }"
+OS_VERSION ?= "${@ str(int('${DATETIME}') - 20160000000000)[:-2] + '0'}"
 OS_VERSION[vardepsexclude] += "DATETIME"
 
 # Ostro OS tries to build minimal images and thus prefers Busybox or
--- a/meta-ostro/conf/distro/ostro.conf
+++ b/meta-ostro/conf/distro/ostro.conf
@@ -236,6 +236,15 @@ PREFERRED_PROVIDER_virtual/java-initial-native = "cacao-initial-native"
 PREFERRED_PROVIDER_virtual/java-native = "cacao-native"
 PREFERRED_PROVIDER_virtual/javac-native = "ecj-bootstrap-native"
 
+# Java runtime: used by packagegroup-java-jdk.bb and packages which
+# RDEPEND on Java.
+#
+# Must use openjdk-8 because Ostro OS is meant to support on-device
+# compilation of Java programs, and we don't want to end up
+# compiling and installing both jre and jdk (they are separate
+# packages which share no files).
+PREFERRED_RPROVIDER_java2-runtime = "openjdk-8"
+
 # OE-core recently switched to gcc 6.1 (OE-core commit c20d863da5700, see tcmode-default.inc).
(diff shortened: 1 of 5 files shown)


pohly:
More commits coming...

pohly:
@Valtis we should be using openjre-8, right?

@kad PR ready for review again

kad:
ok for me.

Valtis:
@pohly We are currently using openjdk-8

meta-ostro/conf/distro/ostro.conf
```diff
@@ -236,6 +236,9 @@ PREFERRED_PROVIDER_virtual/java-initial-native = "cacao-initial-native"
 PREFERRED_PROVIDER_virtual/java-native = "cacao-native"
 PREFERRED_PROVIDER_virtual/javac-native = "ecj-bootstrap-native"
 
+# Java runtime
+PREFERRED_RPROVIDER_java2-runtime = "openjre-8"
```
Valtis: We are currently using openjdk-8: https://github.com/ostroproject/meta-ostro/blob/master/meta-ostro/recipes-core/packagegroups/packagegroup-java-jdk.bb

Whether we actually need to intall javac & friends is arguable however.

pohly: Well, for Java *development* that makes sense. java2-runtime is just used for running Java programs. Does the jdk re-use the jre (i.e. depends on it) or is it something else entirely?

If it depends on it, then java2-runtime = "openjre-8" is fine: we use only that when packagegroup-java-jdk.bb isn't installed, and we add the extra pieces from the jdk when the package group is installed.

Valtis: As far as I can tell, they are separate recipes (no DEPENDS or RDEPENDS on jre in jdk recipe), although both use the same .inc files that hold the common portions of the recipes.

https://github.com/ostroproject/ostro-os/blob/master/meta-java/recipes-core/openjdk/openjdk-8_72b05.bb
https://github.com/ostroproject/ostro-os/blob/master/meta-java/recipes-core/openjdk/openjre-8_72b05.bb

pohly:
@Valtis I don't think we want both openjdk-8 and openjre-8 in the same image, so I changed PREFERRED_RPROVIDER to openjdk-8. Okay? See commit 67d66e1.

@igor-stoppa I had to fix some compatibility issue with Python3 in image-dsk.bbclass. Is commit  	aea4700 okay?

pohly:
@valtis even better, we can also ensure that we do not accidentally include openjre-8 by removing it from the supported recipes. See the revised 2d53c3a.

pohly:
restest this please

meta-ostro/conf/distro/ostro.conf
```diff
@@ -236,6 +236,15 @@ PREFERRED_PROVIDER_virtual/java-initial-native = "cacao-initial-native"
...
 PREFERRED_PROVIDER_virtual/javac-native = "ecj-bootstrap-native"
 
+# Java runtime: pick the same one as in packagegroup-java-jdk.bb.
+#
+# If we pick something simpler here like openjre-8, then images
+# with components like mraa (which depend on java2-runtime) and
+# packagegroup-java-jdk installed pull in both, which then
+# takes up more disk space (openjdk-8 and openjre-8 share no
+# files).
+PREFERRED_RPROVIDER_java2-runtime = "openjdk-8"
```
Valtis: Is there any way we could sync the package group with the PREFERRED_RPROVIDER? Right now, if package group is changed to provide openjre instead, this variable needs to be updated as well, which is something that could be easily forgotten. Thankfully it leads to build breakage rather than silently including both, but it would be nice if the breakage could be avoided as well.

pohly: The package group could be written so that it uses "${PREFERRED_RPROVIDER_java2-runtime}" instead of openjdk-8. If we do that, this comment here needs to be updated.

Shall I make that change?

Valtis: Please do.

Valtis:
Looks good to me.

PR #208, opened 2016-06-13
building-images.rst: add instructions to create .bmap file

gvancuts:
Add instructions on how to generate a bmap file that can be used
by the bmaptool when flashing an Ostro image.

[skip ci]


kad:
better also to add same comments near setting in local.conf.sample

gvancuts:
Good point @kad, I will do this and push a new update

gvancuts:
Updated the PR, rebased and force-pushed.

doc/howtos/building-images.rst
```diff
@@ -207,6 +207,13 @@ Example::
 
 will create both the raw and the VirtualBox appliance images, both compressed.
 
+To get the best benefit out of :command:`bmaptool` when creating your bootable media (see :ref:`booting-and-installation`),
```
dbkinder: How about:  The :command:`bmaptool` works best creating your bootable media (see :ref:`booting-and-instalation`) when the corresponding block map ``.bmap`` file is also generated for the image.  This ``.bmap`` file contains empty block and checksum information about the image that lets  :command:`bmaptool` optimize copying the image to your bootable media and verify what was copied.  To do this ...

kad:
One more instance we can change:
In meta-ostro/classes/ostro-image.bbclass:
OSTRO_VM_IMAGE_TYPES ?= "dsk dsk.ova"
to 
OSTRO_VM_IMAGE_TYPES ?= "dsk dsk.bmap dsk.ova"

gvancuts:
I have just addressed @dbkinder and @kad suggestions and force-pushed a new patch

kad:
👍

PR #210, opened 2016-06-15
conf file updates

pohly:
Support using the files in derived distros better (versioning, cleaner
local.conf.sample with no redundant spaces), more reliable sample
file location tracking.

### DIFF
--- a/meta-ostro/conf/layer.conf
+++ b/meta-ostro/conf/layer.conf
@@ -46,5 +46,5 @@ LAYER_CONF_VERSION = "${OSTRO_OS_LAYER_CONF_VERSION}"
 # conf files which in ostro-os happen to match two files (one from
 # meta and the real one from meta-ostro). Be more specific to avoid
 # user confusion.
-SANITY_LOCALCONF_SAMPLE = "${COREBASE}/meta-ostro/conf/local.conf.sample"
-SANITY_BBLAYERCONF_SAMPLE = "${COREBASE}/meta-ostro/conf/bblayers.conf.sample"
+SANITY_LOCALCONF_SAMPLE = "${META_OSTRO_BASE}/conf/local.conf.sample"
+SANITY_BBLAYERCONF_SAMPLE = "${META_OSTRO_BASE}/conf/bblayers.conf.sample"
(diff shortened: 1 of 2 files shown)


kad:
👍

meta-ostro/conf/layer.conf
```diff
@@ -32,14 +32,19 @@ BBFILE_PRIORITY_ostro = "6"
 # need to be increased by one each time a change is made to
 # local.conf.sample that requires manually updating a local.conf after
 # updating the meta data.
-LOCALCONF_VERSION = "3"
+#
+# A separate, Ostro OS specific variable is used to make the value
+# available to derived distros.
+OSTRO_OS_LOCALCONF_VERSION = "3"
+LOCALCONF_VERSION = "${OSTRO_OS_LOCALCONF_VERSION}"
```
gvancuts: Can you elaborate more on how this would be used by a derivative distro? Depending on how much text is needed, either we can put it in here or make it a paragraph in our on-line doc.

meta-ostro/conf/local.conf.sample
```diff
@@ -2,11 +2,11 @@
 # This file is your local configuration file and is where all local user settings
 # are placed. The comments in this file give some guide to the options a new user
 # to the system might want to change but pretty much any configuration option can
-# be set in this file. More adventurous users can look at local.conf.extended 
+# be set in this file. More adventurous users can look at local.conf.extended
```
gvancuts: Where is this local.conf.extended file?

meta-ostro/conf/layer.conf
pohly: We are currently experimenting with that. I think it is too early to document the approach.

meta-ostro/conf/local.conf.sample
pohly: Seems like we don't have it at all in ostro-os. Might be a Yocto thing. I can remove the entire sentence.

pohly: Removed...

pohly:
@gvancuts okay to merge now?

gvancuts:
:+1:

kad:
Let's merge.

pohly:
@kad okay to merge also without test results? Some builds (https://ostroproject.org/jenkins/job/dsl-meta-ostro_pull-requests/1549/) completed, so it should be safe, but I don't want to take shortcuts without explicit confirmation.

kad:
Let's wait for all builds to complete. Testing should not show surprises for such cosmetic changes, and if builds are all good, we are good to merge.

pohly:
@kad okay, let's wait for build completion. I was a bit confused by your "let's merge"...

PR #211, opened 2016-06-16
image-dsk.bbclass: Replace 'exit 0' with 'return'

Valtis:
Using 'exit 0' prevents any following code from being run. This means
that the unnecessary .dsk file is not being deleted and that .bmap files
are not being generated.


kad:
👍

PR #214, opened 2016-06-22
Add support for OS installer

rojkov:
Since EFI blobs contain predefined PARTUUIDs of rootfs and
are signed we have to generate two different EFI blobs
with different PARTUUIDs. Otherwise the kernel installed
on the device will get confused about which partition to
use as rootfs in case the installation USB stick is still
inserted into the device.

### DIFF
--- a/meta-ostro/classes/ostro-image.bbclass
+++ b/meta-ostro/classes/ostro-image.bbclass
@@ -279,7 +279,7 @@ NOHDD = "1"
 # Image creation: add here the desired value for the PARTUUID of
 # the rootfs. WARNING: any change to this value will trigger a
 # rebuild (and re-sign, if enabled) of the combo EFI application.
-ROOTFS_PARTUUID_VALUE = "12345678-9abc-def0-0fed-cba987654321"
+REMOVABLE_MEDIA_ROOTFS_PARTUUID_VALUE = "12345678-9abc-def0-0fed-cba987654321"
 # The second value is needed for the system installed onto
 # the device's internal storage in order to mount correct rootfs
 # when an installation media is still inserted into the device.
--- a/meta-ostro/classes/ostro-image.bbclass
+++ b/meta-ostro/classes/ostro-image.bbclass
@@ -280,6 +280,10 @@ NOHDD = "1"
 # the rootfs. WARNING: any change to this value will trigger a
 # rebuild (and re-sign, if enabled) of the combo EFI application.
 ROOTFS_PARTUUID_VALUE = "12345678-9abc-def0-0fed-cba987654321"
+# The second value is needed for the system installed onto
+# the device's internal storage in order to mount correct rootfs
+# when an installation media is still inserted into the device.
+INT_STORAGE_ROOTFS_PARTUUID_VALUE = "12345678-9abc-def0-0fed-cba987654320"
 
 # By default, all files will be signed. Once IMA is active and its
 # policy includes a signed file, such signed files can be removed and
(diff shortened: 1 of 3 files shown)


pohly:
@igor-stoppa is this okay to be merged?

meta-ostro/classes/ostro-image.bbclass
```diff
@@ -280,6 +280,10 @@ NOHDD = "1"
 # the rootfs. WARNING: any change to this value will trigger a
 # rebuild (and re-sign, if enabled) of the combo EFI application.
 ROOTFS_PARTUUID_VALUE = "12345678-9abc-def0-0fed-cba987654321"
+# The second value is needed for the system installed onto
+# the device's internal storage in order to mount correct rootfs
+# when an installation media is still inserted into the device.
+ROOTFS_PARTUUID_VALUE2 = "12345678-9abc-def0-0fed-cba987654320"
```
igor-stoppa: I would have choose a slightly more expressive name, like INT_STORAGE_ROOTFS_PARTUUID_VALUE :-)
But apart from that it looks good.
It seems to point to the need for placing the combo files by default elsewhere than in the EFI partition and instead deploy there only the active one. 
@rojkov did you check that the image still fits on a 4GB card, after adding 30MB?

rojkov: In hindsight I see that increasing the efi partitions' size is not really needed since they are populated from the data which can be found in /boot of the rootfs.
Will check if it can be easily avoided.

rojkov:
* reverted boot partition sizes back to 15M;
* renamed `ROOTFS_PARTUUID_VALUE2` to `INT_STORAGE_ROOTFS_PARTUUID_VALUE`;
* rebased, force pushed.

meta-ostro/classes/ostro-image.bbclass
```diff
@@ -280,6 +280,10 @@ NOHDD = "1"
 # the rootfs. WARNING: any change to this value will trigger a
 # rebuild (and re-sign, if enabled) of the combo EFI application.
 ROOTFS_PARTUUID_VALUE = "12345678-9abc-def0-0fed-cba987654321"
+# The second value is needed for the system installed onto
+# the device's internal storage in order to mount correct rootfs
+# when an installation media is still inserted into the device.
+INT_STORAGE_ROOTFS_PARTUUID_VALUE = "12345678-9abc-def0-0fed-cba987654320"
```
igor-stoppa: I promise this is the last nitpicking: how about renaming ROOTFS_PARTUUID_VALUE to something like REMOVABLE_MEDIA_ROOTFS_PARTUUID_VALUE ?

Because that's going to be its sole use, right?

Side comment: I would have preferred to see also the installer, so that the entire use case could be reviewed.
However, for now, apart from the ROOTFS_PARTUUID_VALUE renaming, lgtm.

rojkov:
`ROOTFS_PARTUUID_VALUE` renamed to `REMOVABLE_MEDIA_ROOTFS_PARTUUID_VALUE`.

The installer itself is referenced in this PR if your account has got access to the repo.

igor-stoppa:
@rojkov I do not understand why also the installer could not be part of meta-ostro.

rojkov:
test this please

pohly:
@igor-stoppa I don't have an opinion about this. Is it ready for merging now?

kad:
We can get it to GT first, then move to Ostro

igor-stoppa:
yes, this looks ok

PR #216, opened 2016-06-28
java: Move java behind DISTRO_FEATURES feature

Valtis:
Java is a fairly heavyweight package, and it is likely that at least
some users would like to disable it altogether. Currently both mraa and
upm unconditionally depend on Java, which means it will get included in
the image even if the java-jdk packagegroup is not included.

A new distro feature, java, is introduced to gate both the mraa\upm and
java package group dependencies, so that Java can be easily disabled.

### DIFF
--- a/meta-ostro/classes/ostro-image.bbclass
+++ b/meta-ostro/classes/ostro-image.bbclass
@@ -40,7 +40,6 @@ OSTRO_IMAGE_PKG_FEATURES = " \
     connectivity \
     devkit \
     iotivity \
-    java-jdk \
     nodejs-runtime \
     nodejs-runtime-tools \
     python-runtime \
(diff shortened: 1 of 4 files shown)


pohly:
I was wondering about the BINDINGS_remove because with the "java" distro feature, that line will also remove java in cases where the user explicitly chose to enable java, for example with BINDINGS_pn-upm = "java" in local.conf. But that's a rather contrived scenario, so +1 from me.

kad:
:+1:  from me. I only curious, why not to move those mra/upm to upstream layer ?

Valtis:
I felt that the modifications are too Ostro-specific for upstreaming; if the recipes are used in non-Ostro context, they most likely will not have java in distro feature.

pohly:
Yes, this "java" distro feature is Ostro OS specific.

PR #218, opened 2016-07-01
ostro-image.bbclass: add valgrind to debugging tools

pohly:
valgrind makes (more?) sense as a debugging tool than a profiling tool
(which is where it appears in OE-core), so let's add it to the
"tools-debug" image feature.


pohly:
@rojkov agreed to merge?

rojkov:
Yes, looks good to me

PR #219, opened 2016-07-02
ostro-image.bbclass: install mraa/upm bindings

pohly:
The Java and Python binding packages of mraa and upm were not
installed before because nothing depended on them. The two runtimes
can't depend on them because it is uncertain whether these optional
features are wanted, and the core devkit cannot depend on them because
it is uncertain whether the runtimes are wanted.

At the image level we can decide that based on the IMAGE_FEATURES.

Fixes: IOTOS-1680

### DIFF
--- a/meta-ostro/classes/ostro-image.bbclass
+++ b/meta-ostro/classes/ostro-image.bbclass
@@ -182,7 +182,10 @@ FEATURE_PACKAGES_can = "packagegroup-core-can"
 FEATURE_PACKAGES_ima = "packagegroup-ima-evm-utils"
 
 FEATURE_PACKAGES_iotivity = "packagegroup-iotivity"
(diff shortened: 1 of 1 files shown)


pohly:
@tripzero if you see this before Monday, please review.
@rojkov as it is unsure whether @tripzero will see it soon enough, can you perhaps also check it?

kad:
maybe better to put this to packagegroup-devkit instead of image ? 
Otherwise, this looks ok for me.

pohly:
@kad the packagegroup gets compiled once for all images and cannot change depending on image features, which vary between images.

kad:
ok then

tripzero:
@pohly looks good to me.

pohly:
retest this please

isafw index out of range

PR #220, opened 2016-07-04
add thermal_daemon package

yongli3:
Add userspace thermal daemon utility for thermal management.

Fixes: IOTOS-1675

Signed-off-by: Yong Li <yong.b.li@intel.com>

### DIFF
--- a/meta-ostro/classes/ostro-image.bbclass
+++ b/meta-ostro/classes/ostro-image.bbclass
@@ -152,6 +152,11 @@ IMAGE_INSTALL += "util-linux"
 export ALTERNATIVE_PRIORITY_SHADOW ?= "305"
 IMAGE_INSTALL += "shadow"
 
+# Add Thermal daemon for intel-corei7-64
+IMAGE_INSTALL_THERMAL_DAEMON = ""
+IMAGE_INSTALL_THERMAL_DAEMON_intel-corei7-64 = "thermald"
+IMAGE_INSTALL += "${IMAGE_INSTALL_THERMAL_DAEMON}"
+
 # Additional features and packages used in the base images of
 # ostro-image-swupd.bb and ostro-image-noswupd.bb.
 OSTRO_IMAGE_FEATURES_REFERENCE ?= " \
--- a/meta-ostro/conf/distro/include/ostro-supported-recipes.txt
+++ b/meta-ostro/conf/distro/include/ostro-supported-recipes.txt
@@ -217,6 +217,7 @@ systemd-compat-units@core
 systemd-watchdog@quark-bsp
 systemd@core
 tcp-smack-test@security-smack
+thermald@intel
 tzdata@core
 u-boot-edison@edison
 u-boot-fw-utils-edison@edison


yongli3:
@peijiajames @pohly please help to review. The white list  tool request was submitted: https://whitelisttool.amr.corp.intel.com/view.php?id=2900

pohly:
This PR doesn't actually test anything because the new recipe isn't getting used.

Can you include the rest of the thermald integration here, too?

pohly:
We needed to free up CI resources, so I killed most of the running build jobs.

yongli3:
@pohly I add the tool into packagegroup-tools-interactive, I am not sure if it is the right place for this tool. Please help to review

pohly:
No, the package group is wrong - it's meant to be optional in Ostro OS, while thermald (presumably?) is meant to be included in each Ostro OS image, right?

Does thermald work on all platforms or is it limited to certain specific ones?

The right solution (inside Ostro) would be to add it to IMAGE_INSTALL on all platforms which need it. Something like:
```
IMAGE_INSTALL_THERMAL_DAEMON_intel-corei7-64 = "thermal-daemon"
IMAGE_INSTALL_THERMAL_DAEMON = ""
IMAGE_INSTALL += "${IMAGE_INSTALL_THERMAL_DAEMON}"
```

yongli3:
@pohly Only the corei7-64 is needed. But I found the CI test was building this package in beaglebone platform, how to remove/disable it from beaglebone?

saulwold:
Can you please provide a better DESCRIPTION, your current DESCRIPTION is more like a SUMMARY, also while I know that we (Intel) is supporting this, I am not sure it's exclusively for IA, more of a "Linux Thermal Daemon"

Also, as I noted in the JIRA, we might also need a method to ensure that the correct drivers are loaded on the platform, I am working with Cal Sullivan on a bbclass to ensure that drivers are loaded.

nwang2:
@yongli3 though CI builds the recipe, the final image for non intel-corei7-64 platforms will not include this package.

yongli3:
@saulwold I changed the description to "Linux Thermal Daemon", please check it

saulwold:
@yongli3, great , but you did not address the usage of DESCRIPTION as SUMMARY, what you have for DESCRIPTION is really a SUMMARY = , DESCRIPTION would be a small paragraph about what the recipe provides, while the SUMMARY is a 72 character summary (http://www.yoctoproject.org/docs/1.8/mega-manual/mega-manual.html#var-SUMMARY)

DESCRIPTION http://www.yoctoproject.org/docs/1.8/mega-manual/mega-manual.html#var-DESCRIPTION

Shamelessly Stolen from Steve Sakoman's Thermald Recipe:
SUMMARY = "Linux thermal daemon"

DESCRIPTION = "Thermal Daemon is a Linux daemon used to prevent the \
overheating of platforms. This daemon monitors temperature and applies \
compensation using available cooling methods."


Final question, why GIT recipe instead of using a versioned recipe 1.5.3?

yongli3:
@saulwold Thanks for your comment! Let me update the PR now. Regarding the version, I am thinking the tagged version should be more stable than git version(such as build error, fatal bugs)

saulwold:
@yongli3 your's is OK, since you are adding support for systemd, regarding your comment about using a "tagged" version as being more stable, you should just go ahead and use the version'ed tarball and name the recipe thermald_1.5.3.bb and use the SRC_URI = "https://github.com/01org/thermal_daemon/archive/v${PV}.tar.gz"

As it's clearer that you are using the 1.5.3 version.

You will also need to add the CHKSUMs for that tarball

pohly:
@yongli3 to prevent build issues on platforms where thermal-daemon does not compile or is not needed, add the following to the recipe (picking some Intel MACHINE values here):
```
COMPATIBLE_MACHINE = "intel-corei7-64|intel-core2-32|quark"
```

saulwold:
I agree with Patrick, could you also submit this patch to meta-intel (meta-intel@yoctoproject.org), so that it will be useful to Yocto Project usages also. Please have it in common/recipes-bsp, thanks

yongli3:
Thanks @pohly and @saulwold ! I updated the patch, please help to review. If it is okay, I will send it to meta-intel

nwang2:
@yongli3 Per CI build results, it could also be built on edison. Not sure whether you need to add edison to COMPATIBLE_MACHINE. 
BTW, CI has dropped build on intel-core2-32.

saulwold:
Setting the PV inside the recipe is redundant since PV is derived from the recipe name. Otherwise, it looks good

yongli3:
@saulwold I removed the PV, please  check it again

pohly:
A slightly better recipe (thanks to suggestions from ross) are headed towards meta-intel. I'd prefer to get it merged there and then strip it out of this PR.

yongli3:
@pohly I updated the PR to remove the .bb file since it was merged into meta-intel, please help to review

nwang2:
The entry in supported-recipes list should be thermald@intel.

yongli3:
Thanks @nwang2  I changed it to "@intel"

pohly:
retest this please

pohly:
Now that the recipe is available in meta-intel,  whoever needs it can easily add it themselves also in Ostro OS. Let's keep it out of the base images until we know better what it is needed for.

PR #221, opened 2016-07-04
Enable frequently needed HTTP access components

jlaako:
Enable glib-networking and libsoup recipes which are commonly needed for
various cloud access utilities.


pohly:
The only effect of this change is that these components get included in a world build. Nothing really needs or uses them. I'd prefer to postpone this change until we really have a need for these components.

PR #222, opened 2016-07-05
meta-ostro: Don't enable disk monitoring via BB_DISKMON_DIRS in CI

okartau:
Frequent disk cpace checks create massive amount of transactions
to single NFS server in CI environment, but CI servers disks
are not likely to get full in regular building, so lets avoid
DISKMON checks in CI builds.


okartau:
checked on builders during run of this PR testing, the result was as expected: BB_DISKMON_DIRS was unset, and involved worker made no frequent statfs calls to NFS server. Looking good.

PR #224, opened 2016-07-05
image-dsk.bbclass: add installer param to kernel cmdline

rojkov:
Explicitly add "installer" boot parameter to the kernel command line
used by installation media. This way there's no need to detect if
the media is a USB-stick and as such is supposed to run installer.

The initramfs installed onto the target device's internal
storage shall have the installer disabled always.


pohly:
retest this please

ISA index out of range - a change was merged which should avoid that from now on.

PR #225, opened 2016-07-05
ostro-image.bbclass: Add UPM & mraa nodejs & python bindings

arfoll:
Signed-off-by: Brendan Le Foll <brendan.le.foll@intel.com>

### DIFF
--- a/meta-ostro/classes/ostro-image.bbclass
+++ b/meta-ostro/classes/ostro-image.bbclass
@@ -184,7 +184,8 @@ FEATURE_PACKAGES_ima = "packagegroup-ima-evm-utils"
 FEATURE_PACKAGES_iotivity = "packagegroup-iotivity"
 FEATURE_PACKAGES_devkit = "packagegroup-devkit \
     ${@ bb.utils.contains('IMAGE_FEATURES', 'java-jdk', 'mraa-java upm-java', '', d)} \
-    ${@ bb.utils.contains('IMAGE_FEATURES', 'python-runtime', 'python-mraa', '', d)} \
+    ${@ bb.utils.contains('IMAGE_FEATURES', 'python-runtime', 'python-mraa python-upm', '', d)} \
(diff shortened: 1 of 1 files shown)


meta-ostro/classes/ostro-image.bbclass
```diff
@@ -184,7 +184,8 @@ FEATURE_PACKAGES_ima = "packagegroup-ima-evm-utils"
 FEATURE_PACKAGES_iotivity = "packagegroup-iotivity"
 FEATURE_PACKAGES_devkit = "packagegroup-devkit \
     ${@ bb.utils.contains('IMAGE_FEATURES', 'java-jdk', 'mraa-java upm-java', '', d)} \
-    ${@ bb.utils.contains('IMAGE_FEATURES', 'python-runtime', 'python-mraa', '', d)} \
+    ${@ bb.utils.contains('IMAGE_FEATURES', 'python-runtime', 'python-mraa python-upm', '', d)} \
```
pohly: Does this python-upm depend on some config change? I was not getting such a package when I worked on https://github.com/ostroproject/meta-ostro/pull/219/commits/1f3eb822d9259e122b7bee8eddf990db07723e59, that's why it wasn't added earlier. I thought it was intentional that there is no Python binding for UPM.

arfoll: yes it was but re-enabled with 0.7.2

pohly: That update was merged into ostro-os in parallel to me enabling the packages in the images - that explains why I didn't have it yet.

The commit messages in meta-intel-iot-middlware only mention the version bump, so it wasn't obvious to me that something regarding the bindings might have changed.

Has the root cause of the original problem with the Python bindings, unreliable builds, been identified and fixed?

arfoll: Yes it was an issue in upm's cmake configurations that @pylbert fixed with
the 0.7.2 release. Commits probably should have been split for clarity.

pohly:
retest this please

ISA index out of range error (typically goes away when retrying, under investigation)

PR #228, opened 2016-07-07
Update VirtualBox to latest version

jlaako:
In order to support gcc-5.4, update VirtualBox to latest version.


pohly:
Unrelated test failure and it would have been nice to get more information about the problem (Compile error? Which one? Why do we need this change when apparently it is currently working?), but I guess it can't hurt to merge this.

jlaako:
It is apparently not working if your host compiler is gcc >= 5.4, with following error:
| Checking for gcc:
|   ** gcc version 5.4 found, expected gcc 3.x with x>1 or gcc 4.x with 0<x<10 or gcc 5.3!

PR #229, opened 2016-07-07
add bbappends for jdk and icedtea

tripzero:
This should fix the build issues with the latest meta-java

### DIFF
--- a/meta-ostro-fixes/meta-fixes/recipes-devtools/icedtea/icedtea7-native_%.bbappend
+++ b/meta-ostro-fixes/meta-fixes/recipes-devtools/icedtea/icedtea7-native_%.bbappend
@@ -1,0 +1,7 @@
+
+EXTRA_OEMAKE_append = ' \
+    OE_CFLAGS="${CFLAGS}" \
+    OE_CPPFLAGS="${CPPFLAGS}" \
+    OE_CXXFLAGS="${CXXFLAGS}" \
+    OE_LDFLAGS="${LDFLAGS}" \
+'
--- a/meta-ostro-fixes/meta-fixes/recipes-devtools/openjdk/openjdk-8_%.bbappend
+++ b/meta-ostro-fixes/meta-fixes/recipes-devtools/openjdk/openjdk-8_%.bbappend
@@ -1,0 +1,7 @@
+
+EXTRA_OEMAKE_append = ' \
+    OE_CFLAGS="${CFLAGS}" \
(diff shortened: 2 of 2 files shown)


pohly:
There's no explanation about the problem and why this is the right solution. Also, has upstream been notified? Where? Will they merge the same change (thus making these .bbappends obsolete), or are these permanent distro-specific changes?

Has the problem I mentioned in the bug report with passing target compiler flags to the host compiler when building openjdk-8-native been fixed? Doesn't look like it.

"Fixes: IOTOS-1681" is missing.

pohly:
If the goal is to have some distro-specific fixes while upstream also works on the problem, then meta-ostro/fixes/meta-java/recipes-devtools/ is a better place.

tripzero:
priority one was getting build working, then deal with upstream.  I can do the opposite order if that is more desirable.

pohly:
@tripzero hmm, have you actually tested with the latest upstream repos?

Your test build was with ostro-os master and not the CI branch. Now that I have triggered such a test build (see https://ostroproject.org/jenkins/job/ostro-os_pull-requests/960/ for https://github.com/ostroproject/ostro-os/pull/143) I'm still getting build errors for openjdk.

pohly:
@tripzero regarding dealing with upstream: yes, please engage upstream immediately. That way we (ideally) get this change reviewed before we merge the PR.

tripzero:
"Your test build was with ostro-os master and not the CI branch."

Yes, this test build is with ostro-os master.  I'll test these bbappends against ostro-os to test against CI.

pohly:
@tripzero when you test against ostro-os, make sure that you check out and use the ci-upstream-monitor-master branch. Your previous test build against ostro-os was based on current master and thus would have succeeded also without your changes.

pohly:
@tripzero I couldn't figure out what problems your .bbappends fix - they didn't have any noticable effect in my own tests. FYI, I'm currently testing two fixes on top of the CI branch which allowed me to compile OpenJDK again - see https://github.com/ostroproject/ostro-os/pull/143/commits

tripzero:
@pohly odd.  I was able to produce a failure and a success in my local testing with my patches.

pohly:
@tripzero do we still need these changes (if so, explain why) or can we close the PR?

mythi:
@tripzero what's the need for this?

tripzero:
We can close this

PR #230, opened 2016-07-11
ostro-version.bbclass: introduce single OS_VERSION definition

rojkov:
Since eSDK uses an overriden form for OS_VERSION calculation,
but with wrong input, we'd better have a single definition
for OS_VERSION where we can check the input.

### DIFF
--- a/meta-ostro/classes/ostro-version.bbclass
+++ b/meta-ostro/classes/ostro-version.bbclass
@@ -1,0 +1,57 @@
+# When building in a CI system with swupd enabled, OS_VERSION must be
+# set to a consistent value for all builds. See below
+# for an example how that works with Jenkins.
+#
+# In local builds, we only have ${DATETIME} as something that
+# increments automatically, but it is too large for an integer number
+# on 32 bit systems. Therefore we substract the 2016 as the initial
+# year in which Ostro OS started using swupd and ignore the
+# seconds.
+#
+# The default behavior is to not rebuild just because OS_VERSION
+# changed. If that is desired, include in local.conf:
+#   VERSION_ID = "${OS_VERSION}"
+# This will cause the os-release package to be rebuilt each time
+# OS_VERSION changes, and that in turn causes an image and bundle
+# rebuild.
+#
+# For more predictable results is possible to set OS_RELEASE manually,
+# either in local.conf or in the environment like this:
+#   BB_ENV_EXTRAWHITE="$BB_ENV_EXTRAWHITE OS_VERSION" OS_VERSION=110 bitbake ...
+
+def ostro_get_os_version(d):
+    build_id_str = d.getVar('BUILD_ID', True) or ''
+    if build_id_str:
+        try:
+            build_id = int(build_id_str)
+        except ValueError:
+            # Assume that BUILD_ID is in CI's format, e.g. "2016-07-10_18-49-16-build-127".
+            # We cannot have the definition for OS_VERSION in ostroproject-ci.inc because
+            # the file is used also in eSDKs produced by CI where BUILD_ID is not set in
+            # the expected format.
+            #
+            # In the CI system, the integer OS version used for swupd is derived
+            # from the Jenkins build number part of the BUILD_ID. The conversion
+            # to int and back to string acts as sanity check that we really get a
+            # number out of the BUILD_ID.
+            #
+            # BUILD_ID is expected to have this format:
+            # BUILD_ID = "<parent-build-timestamp (of Jenkins top job)>" + "-build-" + "<jenkins-top-job-build-number>"
+            #
+            # We multiply it by ten, to ensure that the version space has some gap
+            # for minor updates. How to actually build minor updates with Jenkins
+            # still needs to be determined.
+            build_id = int(build_id_str.split('-')[-1]) * 10
+
+        # check if we have reached max value for 32-bit integer
+        if build_id > 2147483646:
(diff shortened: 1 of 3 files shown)


meta-ostro/classes/ostro-version.bbclass
```diff
@@ -0,0 +1,56 @@
...
+        if build_id > 2147483646:
+            bb.warn("The current value for BUILD_ID (%s) doesn't fit 32-bit integer: redefining...")
+            return str(((int(d.getVar('DATETIME', True)) - 20160000000000) / 100) * 10)
+
+        return str(build_id)
+    else:
+        return str(((int(d.getVar('DATETIME', True)) - 20160000000000) / 100) * 10)
+
+OS_VERSION ?= "${@ostro_get_os_version(d)}"
+OS_VERSION[vardepsexclude] += "DATETIME"
```
pohly: I suspect that excluding DATETIME must now happen for ostro_get_os_version and not OS_VERSION. "bitbake-diffsigs" on the sigdata file in stamps for an image should show what kind of dependencies bitbake currently considers.

meta-ostro/classes/ostro-version.bbclass
```diff
@@ -0,0 +1,56 @@
...
+def ostro_get_os_version(d):
+    build_id_str = d.getVar('BUILD_ID', True) or ''
+    if build_id_str:
+        try:
+            build_id = int(build_id_str)
+        except ValueError:
+            # Assume that BUILD_ID is in CI's format, e.g. "2016-07-10_18-49-16-build-127".
+            # We cannot have the definition for OS_VERSION in ostroproject-ci.inc because
+            # the file is used also in eSDKs produced by CI where BUILD_ID is not set in
+            # the expected format.
```
pohly: eSDK should not use ostroproject-ci.inc. That include file is meant exclusively for producing builds inside the CI and besides changing OS_VERSION also has several other undesirable effect (like enabling isafw).

Using the eSDK is more like a local build, and that does not use ostroproject-ci.inc.

Why does the eSDK use ostroproject-ci.inc? That's what we should fix instead of working around it.

meta-ostro/classes/ostro-version.bbclass
rojkov: Yes `bitbake-diffsigs` show that `ostro_get_os_version` still depends on `DATETIME`. Will fix

meta-ostro/classes/ostro-version.bbclass
rojkov: `ostroproject-ci.inc` gets included in the form of `auto.conf` and eSDK includes `auto.conf` explicitly (but doesn't capture the environment used when evaluating `auto.conf`'s expressions).

That's the way how either `ostro-os-production.inc` or `ostro-os-development.inc` end up in eSDK.

pohly: @rojkov can't we exclude auto.conf from the packaged eSDK (but beware of the MACHINE value, see below)? Please check with Paul Eggleton.

Related: we publish four eSDK archives, one per target platform. I suspect that this can be replaced with just a single one where developers then use settings in their own, custom local.conf or auto.conf to choose the actual machine they are interested in. Or do major parts of the cross-compile toolchain already get included in the ostro-x86_64-corei7-64-toolchain-ext-1.0~snapshot.sh archive?

I.e., how target architecture-specific is it really, besides including an auto.conf where the CI has changed the MACHINE value?

rojkov: @bluelightning could you please comment on this?

bluelightning: I'm not sure what mechanism we would use to avoid the eSDK using ostroproject-ci.inc - but if we were to exclude it we'd be getting a substantially different configuration, such that we'd be at risk of sstate artifacts built on the CI no longer matching the configuration used in the eSDK.

My contention on all of this is that it's much more complicated than it needs to be - and the lesser number of changes made by ostroproject-ci.inc, the better. Case in point, the definition of OS_VERSION - simplifying it such that it's set in one place is a good thing to do.

kad: I try to keep it as simple as possible, but people are still pushing more and more stuff in it :)
The issue with that particular case that BUILD_ID is environment variable that is not set by normal user. Putting real value of it inside auto.conf is not something that I'd like to do. Ideally, whatever is inside auto.conf should be ignored for eSDK case  and we should get same sigs. But how to implement that, and more important, how to verify is big open question.

pohly: @bluelightning ostroproject-ci.inc is getting included via auto.conf. Can we exclude the auto.conf from the eSDK?

However, auto.conf also includes the MACHINE setting. When we drop auto.conf, all produced eSDKs will be configured for the same default intel-corei7-64. Can you comment on how different the eSDK content is based on the current MACHINE? Can I install an eSDK produced for intel-corei7-64, change the MACHINE (where?) and then build for another MACHINE?

rojkov:
Excluded `DATETIME` from vardeps of `ostro_get_os_version` and force pushed

rojkov:
* made the type of the build_id variable consistently integer
* force pushed

rojkov:
test this please

pohly:
-1

To  me this approach looks like a workaround for a more fundamental issue with how the generated eSDK is configured. See PR #232 for a different solution.

kad:
This is a hack, but so far the best hack we have. let's go with it. :+1:

PR #232, opened 2016-07-13
CI: simplify auto.conf in eSDK

pohly:
The auto.conf file is used by the CI system to make changes specific
only to the CI. Many of them are intentionally not enabled for
normal development by default, and setting OS_VERSION so that
it uses a CI variable breaks swupd-image creation in the eSDK
because the variable isn't set.

So instead of using the auto.conf created by the CI system, we
create one specifically for the eSDK. Currently two changes are
done this way:
- MACHINE
- configuring images as "development images"

That we have to do the later without explicit user consent
is unfortunate; the original goal of the mechanism was that
users must confirm that they want to build images which aren't
suitable for production. With the eSDK that is currently
not possible.

This is an alternative to PR #230

### DIFF
--- a/meta-ostro/classes/ostroproject-ci.bbclass
+++ b/meta-ostro/classes/ostroproject-ci.bbclass
@@ -1,0 +1,23 @@
+# Used by ostroproject-ci.inc to make some changes which cannot be done
+# directly in the include file.
+
+# The CI uses auto.conf to make some changes which are meant to be active
+# only in the CI, like enabling isafw and buildhistory. None of those
+# are meant to be active when normal developers use the eSDK. At best
+# they cause overhead, but others (like making OS_VERSION depend on
+# a CI variable) outright prevent using the eSDK.
+#
+# Instead of trying to filter the auto.conf, we take full control
+# over its content here.
+python ostroproject_ci_clean_auto () {
+    autoconf = d.expand('${SDK_OUTPUT}/${SDKPATH}/conf/auto.conf')
+    with open(autoconf, 'w') as f:
+        # MACHINE must be set.
+        f.write('MACHINE = "%s"\n' % d.getVar('MACHINE', True))
+        # Currently required because the user cannot add it themselves.
+        # It would be better if the users edited some file to choose
+        # the image mode.
+        f.write('require conf/distro/include/ostro-os-development.inc\n')
+}
+
+SDK_POSTPROCESS_COMMAND_append_task-populate-sdk-ext = "ostroproject_ci_clean_auto; "
--- a/meta-ostro/conf/distro/include/ostroproject-ci.inc
+++ b/meta-ostro/conf/distro/include/ostroproject-ci.inc
@@ -15,6 +15,9 @@ INHERIT += "buildstats-summary"
 # Enable CVE and other security checks.
 INHERIT += "isafw"
 
+# Additional CI specific tweaks.
+INHERIT += "ostroproject-ci"
+
 # Most of the images are expected to contain GPLv3
 # components. Therefore we only enable the license check for those
 # which must not have them (whitelisting), instead of excluding images


pohly:
@bluelightning, @rojkov: this is another way of avoiding the problematic, CI-specific OS_VERSION assigningment, but it also goes beyond that by getting rid of several things that we don't want in the eSDK either (build history, isafw). What do you think?

pohly:
@bluelightning:  is my understanding correct that users of the eSDK have no control over the configuration, i.e. we can't require them to edit some config file before building images?

We use that in our local.conf: https://github.com/ostroproject/meta-ostro/blob/master/meta-ostro/conf/local.conf.sample#L281

bluelightning:
@pohly that is correct - the configuration within the eSDK isn't meant to be changed by the user.

I'm not particularly enthusiastic about doing this. You can try it, but the more things you have in your CI include file and auto.conf used in the CI builds, the more the eSDK configuration is going to diverge from the configuration actually used to build the sstate artifacts. The signatures for all the components that go into the SDK are locked, but that may not be enough to paper over the cracks.

meta-ostro/conf/distro/include/ostroproject-ci.inc
```diff
@@ -15,6 +15,9 @@ INHERIT += "buildstats-summary"
 # Enable CVE and other security checks.
 INHERIT += "isafw"
 
+# Additional CI specific tweaks.
```
kad: I'd call it more "eSDK tweaks" than "CI".

meta-ostro/classes/ostroproject-ci.bbclass
```diff
@@ -0,0 +1,23 @@
...
+# over its content here.
+python ostroproject_ci_clean_auto () {
+    autoconf = d.expand('${SDK_OUTPUT}/${SDKPATH}/conf/auto.conf')
+    with open(autoconf, 'w') as f:
+        # MACHINE must be set.
+        f.write('MACHINE = "%s"\n' % d.getVar('MACHINE', True))
+        # Currently required because the user cannot add it themselves.
+        # It would be better if the users edited some file to choose
+        # the image mode.
+        f.write('require conf/distro/include/ostro-os-development.inc\n')
```
kad: not sure if this is best approach, about just replicating part of content here.

rojkov:
@pohly I don't have strong opinion on what's better. I don't like the fact that the class I've introduced is aware of the very special BUILD_ID format used only in our CI. This is a workaround indeed.

And this PR fixes the problem by dropping a file and creating a new file whereas my gut feeling is that having just the original file must be enough to handle both use cases properly - CI and eSDK - from the beginning. Probably eSDK should ignore `auto.conf` and use a specially crafted `esdk.conf` instead.

But I like isafw and buildhistory don't end up in eSDK with this PR. So I'm inclined to merge this PR instead of #230

pohly:
@rojkov I also like this PR better, but I won't insist. @bluelightning raises a valid concern, and this PR needs more testing to be sure that we don't have sstate issues.

If you feel that this PR is the right way forward, can you do some testing with it? My own testing stopped at "it builds an eSDK with an auto.conf that has the intended content", at which point I wanted to get feedback on the approach. I've not done any real image creation with the generated eSDK.

kad:
to my memory, isafw affects sstate hashes.

rojkov:
In my previous tests updates to auto.conf were ignored by eSDK for some reason. The proper testing would need testing the artifacts made by CI I guess. But I'll try to test it.

rojkov:
The content of auto.conf was not updated in the PR build

http://ostro.ostc.intel.com/download/builds/meta-ostro_pull-requests/232/2016-07-13_09-17-17-build-1595/sdk/intel-corei7-64/ostro-x86_64-corei7-64-toolchain-ext-1.0+snapshot.sh

conf/auto.conf in the eSDK still contains
```
# WARNING: this configuration has been automatically generated and in
# most cases should not be edited. If you need more flexibility than
# this configuration provides, it is strongly suggested that you set
# up a proper instance of the full build system and use that instead.

include conf/distro/include/ostroproject-ci.inc
MACHINE = "intel-corei7-64"
```

kad:
this really looks like a bad hack.

pohly:
@rojkov the content you quoted looks okay to me - setting MACHINE and adding the ostroproject-ci.inc include is exactly what ostroproject_ci_clean_auto() puts into that file. Without this PR, the file has multiple INHERIT statements, irrelevant CI settings and the broken OS_VERSION assignment. I'm just a bit surprised about the header - that's not there when looking at tmp-glibc/work/intel_corei7_64-ostro-linux/ostro-image-noswupd/1.0-r0/sdk-ext/image/usr/local/ostro-x86_64/conf/auto.conf

@kad PR #230 is also a hack. Do you prefer that hack? As I said, I'm okay with "disagree and commit" in this case.

The only clean solution that I see is to configure a CI job with exactly the configuration that we want for the eSDK (no isafw, no buildhistory, normal OS_VERSION) and then generate the eSDK using that configuration.

rojkov:
@pohly ostroproject-ci.inc is not what ostroproject_ci_clean_auto() is supposed to put, but rather ostro-os-development.inc. And swupd-server still fails in the eSDK produced with this PR.

In other tool @bluelightning said that
> eSDK does not "force" the use of those things. The default is to use the same configuration used to build the SDK itself, which usually makes sense. In any event Ostro explicitly sets SDK_INHERIT_BLACKLIST such that buildhistory, isafw, buildstats-summary, archiver and test-iot (among others) get disabled within the eSDK, so those are already being taken care of.

pohly:
@rojkov indeed, conf/distro/include/ostroproject-ci.inc is wrong. I have no idea where that conf/auto.conf comes from that you quoted.

Shall I close this proposal? I don't have time to debug further, and both @kad and @bluelightning had concerns about it.

rojkov:
@pohly Yes, let's close and merge #230.

PR #233, opened 2016-07-14
pseudo: pick pseudo 1.7.5

pohly:
Current automated pulls (https://github.com/ostroproject/ostro-os/pull/116)
run into the pseudo performance regression and moving to OE-core master is
blocked by it.

This PR also needs to be tested with OE-core master before merging,
see https://github.com/ostroproject/ostro-os/pull/144.


pohly:
retest this please

pohly:
@okartau  @incandescant do you agree with this short (?) term workaround until the performance regression is fixed in OE-core?

https://github.com/ostroproject/ostro-os/pull/144 with the same commits completed without running into the performance problem, which shows that they work as intended.

incandescant:
@pohly I think this workaround makes sense, yes. Better to carry an older pseudo and otherwise track master than fall too far behind.

okartau:
yes, makes sense

PR #234, opened 2016-07-15
ostro-version.bbclass: Python3-compatible OS_VERSION

pohly:
The initial ostro-version.bbclass reverted to the Python3-incompatible
expression where  / 100 results in a floating point string.

Found when trying the code with Python3,see https://ostroproject.org/jenkins/job/build-recipe-selftests/2561/console


pohly:
The commit is also getting tested as part of https://github.com/ostroproject/ostro-os/pull/148 (besides the manual testing that I did before). Let's merge now, to avoid loosing more time.

PR #235, opened 2016-07-15
Backport OE-Core CVE fixes

avalluri:
Back-ported the oe-core package upgrades that have fixed CVEs.
This is a temporary fix, should be reverted once we move to oe-core master.


avalluri:
Updated: Fix the build failures on Intel platforms.

pohly:
@avalluri thanks for helping with plan B, the backporting. But plan A, updating to Ostro OS, seems to work out, so let's follow plan A.

avalluri:
@pohly , Note that dosfstools needs to be patched with an upstream fix to work properly for configuration.

PR #236, opened 2016-07-16
linux-yocto_4.4.bbappend: freeze intel-quark kernel

pohly:
Updating meta-intel breaks compilation for intel-quark.
To avoid that build breakage until
https://bugzilla.yoctoproject.org/show_bug.cgi?id=9961
gets sorted out we freeze the intel-quark
kernel at the previous version.

This is relevant to avoid build breakages together
with the update meta-intel, see https://github.com/ostroproject/ostro-os/pull/147


pohly:
@rojkov do you agree that we should work around the problem like this?

rojkov:
@pohly Yes, better to resort to the last known good kernel.

PR #237, opened 2016-07-18
ostro-version.bbclass: leave only one special case check

rojkov:
Modify ostro_get_os_version to return the default value
based on current date/time always except one CI specific
use case.


pohly:
It would have been good to mention that this was motiviated by a real problem with the previous approach, the "WARNING: The current value for BUILD_ID (20160718124406) doesn't fit 32-bit integer: redefining..." warnings that one gets during parsing.

Other than that, +1 from me.

rojkov:
Amended commit message and force pushed.

PR #238, opened 2016-07-18
image-dsk.bbclass: deploy uefiapp into per-image directories

pohly:
This should fix the race conditions that we have seen in some internal builds.

@igor-stoppa okay to merge once it has passed automated CI testing?


pohly:
Note that none of the usual image file handling gets applied, i.e. no symlink, no removal of old files. This would also have to be added later, if desired.

pohly:
@igor-stoppa pointed out that creating the uefiapp should not be done once per bundle, and I agree. This is something that should be fixed separately, though, and not hold up fixing the race condition and the resulting build breakages.

pohly:
Merged as discussed yesterday with Igor on IRC.

PR #239, opened 2016-07-19
ostro: switch to latest swupd 3.6.0

pohly:
@igor-stoppa: we need to start building in the CI system now to get
this into the snapshot today. Therefore I went ahead and prepared this
PR.

I discussed the upcoming "enforce signature check unless explicitly
disable" patch with @ereshetova and she agreed that it is okay to use
3.6.0 without that patch for now, as long as we add that patch as soon
as it is available.

### DIFF
--- a/meta-ostro/conf/distro/include/ostro-supported-recipes.txt
+++ b/meta-ostro/conf/distro/include/ostro-supported-recipes.txt
@@ -49,6 +49,7 @@ bison@core
 bluetooth-rfkill-event@edison
 bluez5@core
 boost@core
+bsdiff@meta-swupd
 busybox@core
 bzip2@core
 ca-certificates@core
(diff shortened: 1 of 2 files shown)


pohly:
Don't merge yet, according to @igor-stoppa it lacks some patch that needs to go into a .bbappend file.

igor-stoppa:
@pohly :

(14:51:35) igor: as i said, it can get in

pohly:
@igor-stoppa please submit your PR with the patch and then we can decide.

pohly:
The PR including the EFI updater change is in https://github.com/ostroproject/ostro-os/pull/150

pohly:
Clarified with Igor that we should take this PR and merge the EFI update change later.

PR #244, opened 2016-07-22
local.conf: disable packaging of source code

pohly:
I tested this mode locally before creating the PR. The PR itself
still needs to be verified (i.e. we need to look at the content
of the resulting images).

### DIFF
--- a/meta-ostro/conf/layer.conf
+++ b/meta-ostro/conf/layer.conf
@@ -35,7 +35,7 @@ BBFILE_PRIORITY_ostro = "6"
 #
 # A separate, Ostro OS specific variable is used to make the value
 # available to derived distros.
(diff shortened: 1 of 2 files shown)


pohly:
Content looks good now (from ostro-image-swupd-all-intel-corei7-64.dsk in build-1612):
```
$ tree -a mnt/usr/src/
mnt/usr/src/

0 directories, 0 files

$ ls -l mnt/lib/.debug/libc-2.23.90.so 
-rwxr-xr-x 1 root root 9593136 Jul 22 11:13 mnt/lib/.debug/libc-2.23.90.so
```

pohly:
Merging despite the unrelated Edison test failure.

PR #249, opened 2016-08-09
default-ip6tables.rules: Update ipv6 firewall rules

skuusela:
Pythons unit test (test_nntplib.py) fails because firewall doesn't
allow ipv6 input to localhost. Adding a rule from iptables.rules to
ip6tables.rules fixes this.

Signed-off-by: Simo Kuusela <simo.kuusela@intel.com>

### DIFF
--- a/meta-ostro/recipes-security/iptables-settings-default/files/default-ip6tables.rules
+++ b/meta-ostro/recipes-security/iptables-settings-default/files/default-ip6tables.rules
@@ -7,6 +7,7 @@
 # allow containers to access DHCP service
 -A INPUT -i ve-+ -p udp -m udp --dport 67 -j ACCEPT
 -A INPUT -s fe80::/10 -j ACCEPT
+-A INPUT -i lo -j ACCEPT
 -A INPUT -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
 -A INPUT -p ipv6-icmp -j ACCEPT
 # allow forwarding traffic to/from containers
--- a/meta-ostro/recipes-security/iptables-settings-default/files/default-ip6tables.rules
+++ b/meta-ostro/recipes-security/iptables-settings-default/files/default-ip6tables.rules
@@ -4,10 +4,9 @@
 :OUTPUT ACCEPT
 # allow containers to access DNS service
 -A INPUT -i ve-+ -p udp -m udp --dport 53 -j ACCEPT
-# allow containers to access DHCP service
--A INPUT -i ve-+ -p udp -m udp --dport 67 -j ACCEPT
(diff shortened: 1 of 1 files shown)


ipuustin:
Good catch. The purpose of the default firewall rules has always been to allow loopback access:
> Briefly, all incoming packets are dropped, except for those belonging to already established connections or those that are coming from the loopback interface.

Looking at the rules, it seems that there's a thinko (everything from link-local addresses is allowed, which is IMHO too wide since ICMPv6 is anyway allowed later). Could you remove the `-A INPUT -s fe80::/10 -j ACCEPT` line in a separate patch too and add it to this PR?

skuusela:
test this please

skuusela:
test this please

ipuustin:
Apparently Iotivity tests fail? It looks like the test case only handles IPv4 firewall rules. This patch to https://github.com/ostroproject/meta-iotqa might help open the corresponding ports on IPv6 side too.

```
diff --git a/lib/oeqa/runtime/iotivity/iotvt_integration.py b/lib/oeqa/runtime/iotivity/iotvt_integration.py
index 45d4f73..ec70def 100644
--- a/lib/oeqa/runtime/iotivity/iotvt_integration.py
+++ b/lib/oeqa/runtime/iotivity/iotvt_integration.py
@@ -38,7 +38,9 @@ class IOtvtIntegration(oeRuntimeTest):
         cls.tc.target.run("killall simpleclientserver threadingsample")
         # Setup firewall accept for multicast
         cls.tc.target.run("/usr/sbin/iptables -w -A INPUT -p udp --dport 5683 -j ACCEPT")
+        cls.tc.target.run("/usr/sbin/ip6tables -w -A INPUT -p udp --dport 5683 -j ACCEPT")
         cls.tc.target.run("/usr/sbin/iptables -w -A INPUT -p udp --dport 5684 -j ACCEPT")
+        cls.tc.target.run("/usr/sbin/ip6tables -w -A INPUT -p udp --dport 5684 -j ACCEPT")
 
     def presence_check(self, para):
         '''this is a function used by presence test
```

@zhangjingke, what do you think? Do the iotivity tests require any other ports to be opened? https://github.com/solettaproject/soletta/issues/1380 has some discussion about a "random port to receive unicast", do you know how that port is assigned?

ipuustin:
Opened https://github.com/ostroproject/meta-iotqa/pull/194 so that Iotivity tests can open required ports.

skuusela:
test this please

mythi:
retest this please

mythi:
@skuusela @ipuustin this is good now. the meta-iotqa changes were merged in master

PR #250, opened 2016-08-11
Add network dependency in timesyncd

neofang7:
Fixes: IOTOS-1743

During Ostro shutdown, systemd-timesyncd service will start up to sync
with time server. However, network services are very possible to be
stopped before timesyncd starts up. So, add network.target dependency to
timesyncd.service to garantee that timesyncd will be triggered before
network services stop.

Signed-off-by: neofang7 <neo.fang@intel.com>


mythi:
@neofang7 What's the Upstream-Status to this?

neofang7:
@mythi @peijiajames @nwang2 
Please help to review.

neofang7:
@mythi This patch is only to improve Ostro shutdown performance on GT boards.
Is it necessary to deliver to upstream?

mythi:
@neofang7 to me it looks you are fixing an obvious upstream race condition (nothing Ostro specific).

mythi:
@neofang7 any update?

mythi:
@neofang7 any update?

jlaako:
Many systemd services, even the ones shipped with systemd itself have similar faults. However, you may want to use network-online.target instead, because for interfaces using DHCP the network.target is reached already before DHCP has successfully obtained an IP address...

PR #254, opened 2016-08-23
openjdk-8_%.bbappend: Add fix for the openjdk QA error

Valtis:
Openjdk fails the QA check, as some of the makefiles ignore the
EXTRA_LDFLAGS variable, which is used to pass the Yocto linker settings.
This causes the build to fail due to a bad --hash-style.

Offending makefiles are patched to use the variable to fix the error


mythi:
@Valtis thanks! Did you test this with http://git.yoctoproject.org/cgit.cgi/poky/commit/?id=a98a8180863ff45b477a1f8439ebcec21151d282

Valtis:
Yes, I updated my oe-core & bitbake dependencies before testing anything. Build fails without patch, succeeds with patch.

mythi:
Not needed anymore since the fix is in meta-java. @Valtis thanks!

PR #255, opened 2016-08-25
iotsstatetests.py: python3: convert stdout byte stream to str

mythi:
The subprocess.PIPE as stdout returns byte stream and thus cannot be
used with str.join().

Convert the byte stream to str to avoid a TypeError with str.join():
"TypeError: sequence item 0: expected str instance, bytes found"

### DIFF
--- a/meta-ostro/lib/oeqa/selftest/iotsstatetests.py
+++ b/meta-ostro/lib/oeqa/selftest/iotsstatetests.py
@@ -127,7 +127,7 @@ MACHINE = \"%s\"
                                      stdout=subprocess.PIPE,
                                      stderr=subprocess.STDOUT)
                 stdout, stderr = p.communicate()
-                analysis.append(stdout)
+                analysis.append(stdout.decode("utf-8"))
         if len(errors) > 1:
             # If this fails, it often fails for a whole range of tasks where one depends on
             # the other. In this example, only the original source file was different:


meta-ostro/lib/oeqa/selftest/iotsstatetests.py
```diff
@@ -127,7 +127,7 @@ def get_hashes(d, subdir):
                                      stdout=subprocess.PIPE,
                                      stderr=subprocess.STDOUT)
                 stdout, stderr = p.communicate()
-                analysis.append(stdout)
+                analysis.append(stdout.decode("ascii"))
```
pohly: I prefer decode("utf-8") - a bit more future-proof, but shouldn't matter in practice.

mythi: On Thu, Aug 25, 2016 at 9:26 AM, Patrick Ohly <notifications@github.com>
wrote:

> In meta-ostro/lib/oeqa/selftest/iotsstatetests.py
> <https://github.com/ostroproject/meta-ostro/pull/255#discussion_r76186611>
> :
>
> > @@ -127,7 +127,7 @@ def get_hashes(d, subdir):
> >                                       stdout=subprocess.PIPE,
> >                                       stderr=subprocess.STDOUT)
> >                  stdout, stderr = p.communicate()
> > -                analysis.append(stdout)
> > +                analysis.append(stdout.decode("ascii"))
>
> I prefer decode("utf-8") - a bit more future-proof, but shouldn't matter
> in practice.
>

I thought about it. Will change to utf-8.

mythi:
v2: using utf-8

pohly:
+1

PR #256, opened 2016-08-25
ostro-image-swupd.bb: Add some AppFW sample apps in qa-bundle-b

daweiwu:
To verify IOTOS-1672 in Ostro OS build , we need one bundle which
includes AppFW app .

Signed-off-by: Wu Dawei <daweix.wu@intel.com>


meta-ostro/recipes-image/images/ostro-image-swupd.bb
```diff
@@ -85,6 +85,10 @@ BUNDLE_CONTENTS[qa-bundle-a] = " \
 BUNDLE_CONTENTS[qa-bundle-b] = " \
     hello-bundle-b \
     hello-bundle-s \
+    example-app-c  \
```
mythi: pick just one.

daweiwu:
updated. Just one now.

PR #262, opened 2016-11-30
swupd enhancements

pohly:
These are changes from
https://github.com/ostroproject/ostro-os/pull/198 which can resp. must
(the swupd distro config change) be merged before also updating
meta-swupd.


pohly:
@mythi, @ipuustin: can you help review and merge this? Automated testing has passed fine.

I'd like to update Ostro OS to the improved meta-swupd. Is that okay?

mythi:
@pohly I've already reviewed the changes and they look OK. Thanks!

meta-ostro/conf/distro/ostro.conf
```diff
@@ -22,6 +22,20 @@ SDK_UPDATE_URL = "${OSTRO_SDK_UPDATE_URL}"
...
+# The information about where to find version information and actual
+# content is needed in several places:
+# - the swupd client in the image gets configured such that it uses that as default
+# - swupd server needs information about the previous build
+#
+# The version URL determines what the client picks as the version that it updates to.
+# The content URL must have all builds ever produced and is expected to also
+# have the corresponding version information.
+SWUPD_VERSION_URL ?= "https://download.ostroproject.org/updates/ostro-os/milestone/${MACHINE}/${SWUPD_IMAGE_PN}"
+SWUPD_CONTENT_URL ?= "https://download.ostroproject.org/updates/ostro-os/builds/${MACHINE}/${SWUPD_IMAGE_PN}"
```
kad: builds/ directory is something that can be removed at any point. only hierarchy over milestone/ is known to be preserved always, so content url also should be pointing to the place where version is taken from.

pohly: @kad what you describe is the current deployment model. In this model, doing incremental builds (= reusing results from a previous build) is impossible because it isn't guaranteed that the content of the previous build gets published. It happened to work only because support for incremental builds was missing, which (as a side effect) made each build self-contained.

This should be changed because it is inefficient. I know it has implications for the Ostro CI and release process, which needs to be discussed should we ever do another release...

pohly:
mythi: do not merge yet. I'm going to try a build against the latest milestone (SWUPD_CONTENT_URL pointing to the milestone directory), to see how much slower that is.

mythi:
@pohly ok

pohly:
I've added another commit which removes usage of the intermediate "builds" directory. As expected, building became slower again (64 minutes instead of 53), but it's nevertheless the right approach for the current release procedure. I've intentionally added a separate commit so that I could record the reasoning behind the change.

One more meta comment: I found it a bit surprising that the "builds" were only meant for testing swupd and not guaranteed to be around. I thought the goal was to have people follow the latest master, which implies that they continuously update their test devices to the latest build, not just the latest milestone. I know that this hasn't worked in practice, but that just means we hadn't reached that goal.

PR #264, opened 2016-12-15
ostro-version.bbclass: bump OS_VERSION by 1000000000 in non-master CI builds

pohly:
OS_VERSION must not go backwards, which was not guaranteed when
deriving OS_VERSION from the BUILD_ID in pull request or component
master builds, because those builds have a different numbering than
the production builds for the ostro-os master.

Now we detect the kind of build and bump the OS_VERSION by 1000000000
in those non-master builds. Depends on support in the CI system for
CI_BUILD_TYPE.


okartau:
test this please

okartau:
now that builder exporting CI_BUILD_TYPE was merged: test this please

